PR #7, opened 2015-03-09
fix: objc native body condition and test describes

thibaultcha:


ahmadnassri:
@thibaultCha needs rebasing, and should now use new test fixtures (automatically tested, just need to create the output files)

ahmadnassri:
see: `test/fixtures/output` and `test/fixtures/requests` for reference

ahmadnassri:
also, don't forget to update `README.md` with the table of options.

PR #14, opened 2015-03-10
Fix: objc/native tests + enhancements

thibaultcha:
- `objc/native` timeout option (will be documented soon)
- `objc/native` fixtures + fix the generation
- missing fixture file results in failing test (for the fixture) (this
  behaviour is the same as before) but prints a nicer failure message
- if the fixture file is missing, snippet test will be set to pending to
  avoid a double failure
- remove done() calls in tests, we are not testing anything asynchronous
- improved applciation-form-encoded.json fixture with a new parameter

### DIFF
--- a/src/targets/objc/native.js
+++ b/src/targets/objc/native.js
@@ -28,7 +28,7 @@ module.exports = function (options) {
   }
 
   // Set request body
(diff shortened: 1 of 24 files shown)


thibaultcha:
Closes #7

PR #16, opened 2015-03-10
objc: rewrote native with improved syntax

thibaultcha:
- objc/native now creates NSDictionary literals for headers and
  parameters if possible
- objc/native has a `timeout`, `indent` and `pretty` options
- this makes helpers.js a valid file to put in a family folder (ignored
  when loading all available targets)

I think that every language should be allowed to parse the JSON if
postData.text is a valid JSON object, allowing the developer to have
more flexibility over the snippet rather than a huge JSON.stringify
string:

``` objective-c
#import <Foundation/Foundation.h>

NSDictionary *headers = @{ @"Content-Type": @"application/json" };
NSDictionary *parameters = @{ @"number": @1,
                              @"string": @"f\"oo",
                              @"arr": @[ @1, @2, @3 ],
                              @"nested": @{ @"a": @"b" },
                              @"arr_mix": @[ @1, @"a", @{ @"arr_mix_nested": @{  } } ] };

NSData *postData = [NSJSONSerialization dataWithJSONObject:parameters options:0 error:nil];

NSMutableURLRequest *request = [NSMutableURLRequest requestWithURL:[NSURL URLWithString:@"http://mockbin.com/har"]
                                                       cachePolicy:NSURLRequestUseProtocolCachePolicy
                                                   timeoutInterval:10.0];
[request setHTTPMethod:@"POST"];
[request setAllHTTPHeaderFields:headers];
[request setHTTPBody:postData];

NSURLSession *session = [NSURLSession sharedSession];
NSURLSessionDataTask *dataTask = [session dataTaskWithRequest:request
                                            completionHandler:^(NSData *data, NSURLResponse *response, NSError *error) {

}];

[dataTask resume];
```

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -1,66 +1,66 @@
-'use strict';
+'use strict'
 
-var debug = require('debug')('httpsnippet');
-var es = require('event-stream');
-var FormData = require('form-data');
-var qs = require('querystring');
-var reducer = require('./reducer');
-var targets = require('./targets');
-var url = require('url');
-var util = require('util');
-var validate = require('har-validator');
+var debug = require('debug')('httpsnippet')
+var es = require('event-stream')
+var FormData = require('form-data')
+var qs = require('querystring')
+var reducer = require('./reducer')
+var targets = require('./targets')
+var url = require('url')
+var util = require('util')
+var validate = require('har-validator')
 
 // constructor
 var HTTPSnippet = function (req, lang) {
-  this.source = util._extend({}, req);
+  this.source = util._extend({}, req)
 
   // add optional properties to make validation successful
-  this.source.httpVersion = this.source.httpVersion || 'HTTP/1.1';
-  this.source.queryString = this.source.queryString || [];
-  this.source.headers = this.source.headers || [];
-  this.source.cookies = this.source.cookies || [];
-  this.source.postData = this.source.postData || {};
-  this.source.postData.mimeType = this.source.postData.mimeType || 'application/octet-stream';
+  this.source.httpVersion = this.source.httpVersion || 'HTTP/1.1'
+  this.source.queryString = this.source.queryString || []
+  this.source.headers = this.source.headers || []
+  this.source.cookies = this.source.cookies || []
+  this.source.postData = this.source.postData || {}
+  this.source.postData.mimeType = this.source.postData.mimeType || 'application/octet-stream'
 
-  this.source.bodySize = 0;
-  this.source.headersSize = 0;
-  this.source.postData.size = 0;
+  this.source.bodySize = 0
+  this.source.headersSize = 0
+  this.source.postData.size = 0
 
   validate.request(this.source, function (err, valid) {
     if (!valid) {
-      throw err;
+      throw err
     }
 
     // construct query string object
-    this.source.queryObj = {};
-    this.source.headersObj = {};
-    this.source.allHeaders = {};
-    this.source.postData.jsonObj = false;
-    this.source.postData.paramsObj = false;
+    this.source.queryObj = {}
+    this.source.headersObj = {}
+    this.source.allHeaders = {}
+    this.source.postData.jsonObj = false
+    this.source.postData.paramsObj = false
 
     // construct query objects
     if (this.source.queryString && this.source.queryString.length) {
-      debug('queryString found, constructing queryString pair map');
+      debug('queryString found, constructing queryString pair map')
 
-      this.source.queryObj = this.source.queryString.reduce(reducer, {});
+      this.source.queryObj = this.source.queryString.reduce(reducer, {})
     }
 
     // construct headers objects
     if (this.source.headers && this.source.headers.length) {
       // loweCase header keys
       this.source.headersObj = this.source.headers.reduceRight(function (headers, header) {
-        headers[header.name.toLowerCase()] = header.value;
-        return headers;
-      }, {});
+        headers[header.name.toLowerCase()] = header.value
+        return headers
+      }, {})
     }
 
     // construct Cookie header
     var cookies = this.source.cookies.map(function (cookie) {
-      return encodeURIComponent(cookie.name) + '=' + encodeURIComponent(cookie.value);
-    });
+      return encodeURIComponent(cookie.name) + '=' + encodeURIComponent(cookie.value)
+    })
 
     if (cookies.length) {
-      this.source.allHeaders.cookie = cookies.join('; ');
+      this.source.allHeaders.cookie = cookies.join('; ')
     }
 
     switch (this.source.postData.mimeType) {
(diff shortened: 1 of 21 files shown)


src/index.js
```diff
@@ -113,7 +113,7 @@ module.exports.availableTargets = function () {
     var clients = Object.keys(targets[key])
 
     .filter(function (prop) {
-      return !~['info', 'index'].indexOf(prop);
+      return !~['info', 'index', 'helpers'].indexOf(prop);
```
ahmadnassri: this should be redundant, since you're excluding the file in `index.js`

thibaultcha: Mhh yeah probably, I did it backwards (currently rebasing)

src/targets/objc/native.js
```diff
@@ -12,48 +14,74 @@ module.exports = function (options) {
   // Dependencies
   code.push('#import <Foundation/Foundation.h>');
   code.push(null);
-  code.push('NSURLSession *session = [NSURLSession sharedSession];');
-  code.push(null);
+
+  // Build headers (including cookies)
+  var headers = this.source.headers.reduce(function (headers, header) {
```
ahmadnassri: there's a helper object thats already defined: `this.source.headersObj`

src/targets/objc/native.js
```diff
@@ -12,48 +14,74 @@ module.exports = function (options) {
...
+  }
+
+  if (Object.keys(headers).length) {
+    code.push(objcHelpers.nsDictionaryBuilder('headers', headers, opts.indent));
+  }
+
+  var parameters;
+
+  // Form-encoded body
+  if (this.source.postData.params && this.source.postData.params.length) {
```
ahmadnassri: this is wrong, you need to check for `mimeType` here since it will be the same object for multipart

test/index.js
```diff
@@ -8,6 +8,7 @@ describe('HTTPSnippet', function () {
   var targets = HTTPSnippet.availableTargets();
 
   targets.map(function (target) {
+    console.log(target)
```
ahmadnassri: ^

src/targets/objc/native.js
```diff
@@ -1,65 +1,84 @@
...
+  }
+
+  var parameters;
+
+  if (this.source.postData.mimeType && (this.source.postData.text || this.source.postData.params)) {
+    switch (this.source.postData.mimeType) {
+      case 'application/x-www-form-urlencoded':
+        parameters = this.source.postData.params;
+
+        code.push(null);
```
thibaultcha: @ahmadnassri As already explained before, i have no choice but to do that, otherwise, an example: `headers.json`

``` js
{
  "method": "GET",
  "url": "http://mockbin.com/har",
  "headers": [
    {
      "name": "Accept",
      "value": "application/json"
    },
    {
      "name": "X-Foo",
      "value": "Bar"
    }
  ]
}
```

Will enter the switch, and create a line:

``` objective-c
NSData *postData = [[NSData alloc] initWithData:[@"undefined" dataUsingEncoding:NSUTF8StringEncoding]];
```

And actually, since I just updated the `plain/text` case with a `v.replace(/"/g, '\\"')`, we will now have an undefined error.

src/targets/objc/native.js
```diff
@@ -1,65 +1,84 @@
...
+        }
+        break;
+
+      case 'application/json':
+        try {
+          parameters = JSON.parse(this.source.postData.text);
+          code.push(objcHelpers.nsDictionaryBuilder('parameters', parameters, opts.indent));
+          code.push(null);
+          code.push('NSData *postData = [NSJSONSerialization dataWithJSONObject:parameters options:0 error:nil];')
+        } catch (e) {}
```
thibaultcha: Not sure how to handle this here.

src/targets/objc/native.js
```diff
@@ -1,65 +1,84 @@
...
+        break;
+
+      case 'multipart/form-data':
+        break;
+
+      default:
+        code.push(null);
+        code.push('NSData *postData = [[NSData alloc] initWithData:[@"' + this.source.postData.text.replace(/"/g, '\\"') + '" dataUsingEncoding:NSUTF8StringEncoding]];');
```
ahmadnassri: wouldn't it be safer here to apply `JSON.stringify` rather than running a `str.replace`?

src/targets/objc/native.js
ahmadnassri: `this.source.postData.mimeType` is always set as per [`src/index.js#L21`](https://github.com/ahmadnassri/httpsnippet/blob/master/src/index.js#L21), defaults to application/octet-stream.

and since I'm doing the detection and parsing of the postData in [`src/index.js#L38`](https://github.com/ahmadnassri/httpsnippet/blob/master/src/index.js#L38) you can rely on it taking care of all parsing scenarios and just inspect for mimeType values you want to handle.

src/targets/objc/native.js
ahmadnassri: I'm dealing with this here: [`src/index.js#L50`](https://github.com/ahmadnassri/httpsnippet/blob/master/src/index.js#L50)

basically, attempt to parse and if no joy, keep it as string and change the switch to `text/plain`.

src/targets/objc/native.js
ahmadnassri: I might be wrong ... will have to test different scenarios ... like what if the text was a JSON string already ...

src/targets/objc/native.js
```diff
@@ -1,57 +1,76 @@
...
+      case 'application/json':
+        try {
+          parameters = JSON.parse(source.postData.text);
+          code.push(objcHelpers.nsDictionaryBuilder('parameters', parameters, opts.indent));
+          code.push(null);
+          code.push('NSData *postData = [NSJSONSerialization dataWithJSONObject:parameters options:0 error:nil];')
+        } catch (e) {}
+        break;
+
+      case 'multipart/form-data':
```
ahmadnassri, line 51: if you're not adding any special conditions for file management / multipart params, remove this, and the fall back will always work in sending `source.postData.text` directly ...

src/targets/objc/native.js
```diff
@@ -1,57 +1,76 @@
...
-      code.push('NSMutableData *postData = [[NSMutableData alloc] initWithData:[@"' + params[0].name + '=' + params[0].value + '" dataUsingEncoding:NSUTF8StringEncoding]];');
-      for (var i = 1, len = params.length; i < len; i++) {
-        code.push('[postData appendData:[@"&' + params[i].name + '=' + params[i].value + '" dataUsingEncoding:NSUTF8StringEncoding]];');
-      }
-    } else if (source.postData.mimeType === 'application/json' && source.postData.text) {
-      code.push('NSData *postData = [[NSData alloc] initWithData:[@' +  JSON.stringify(source.postData.text) + ' dataUsingEncoding:NSUTF8StringEncoding]];');
-    } else if (source.postData.text) {
-      code.push('NSData *postData = [[NSData alloc] initWithData:[@"' + source.postData.text + '" dataUsingEncoding:NSUTF8StringEncoding]];');
-    }
+  if (Object.keys(source.headersObj).length) {
```
ahmadnassri: you're using `source.allHeaders` best to keep these identical (use-case: no `source.headers`, but cookie was added to `source.allHeaders`)

src/targets/objc/native.js
```diff
@@ -1,57 +1,76 @@
...
 
-  // Dependencies
   code.push('#import <Foundation/Foundation.h>');
+
+  if (Object.keys(source.allHeaders).length) {
+    code.push(null);
+    code.push(objcHelpers.nsDictionaryBuilder('headers', source.allHeaders, opts.indent));
+  }
+
+  var parameters;
```
ahmadnassri: add default value to `false` maybe here, that if statement for this later looks wrong without it.

test/fixtures/output/curl/application-json.sh
```diff
@@ -1,4 +1,4 @@
 curl --request POST \
   --url "http://mockbin.com/har" \
-  --header "content-type: application/json" \
-  --data "{\"foo\": \"bar\"}"
+  --header "Content-Type: application/json" \
```
ahmadnassri: please rebase + regenerate, all current output fixtures are functional and working, this will break them (headers are lower cased now)

src/targets/objc/native.js
thibaultCha: Well, now yes.

thibaultcha, line 51: 1. text used to not be set in that case
2. I used that option because it's simpler for developers to edit/remove a parameter once pasted. So really not sure.

src/targets/objc/native.js
thibaultcha: same, that's new, need to update

src/targets/objc/native.js
thibaultcha: It's an object. It's declared as undefined. If there's something, it'll be an object. Seems better than declaring a boolean and later putting an object into it.

ahmadnassri: yeah, I could see that too..

thibaultcha: I'd rather check for undefined in the `if` condition. Anyways this might even disappear now with the new helpers.

thibaultcha:
1

ahmadnassri:
:clap:

Issue #12, opened 2015-03-10
Add node > unirest option


therebelrobot:
(again, adding to get an issue #)

ahmadnassri:
aaaaand done!

PR #18, opened 2015-03-11
[#12][feature] Added unirest (multipart is still failing)

therebelrobot:
I left in the request ones here to show in mocha that they're broken. The unirest documentation says multipart is experimental, and I couldn't find enough examples to make an adequate snippet for it.


sonicaghi:
+1

PR #20, opened 2015-03-11
tests: stronger application-json.json fixture

thibaultcha:
Losing my mind on every rebase I'm doing with that, so here's a PR just for this.

This allows languages willing to parse the JSON object to test the behaviour against a nastier JSON object. Targets sending a string should see no difference.


ahmadnassri:
:+1:

PR #22, opened 2015-03-11
Java module

shashiranjan84:
Added java target using Unirest api.

### DIFF
--- a/src/targets/java/index.js
+++ b/src/targets/java/index.js
@@ -1,0 +1,5 @@
+'use strict';
+
+module.exports = require('require-directory')(module);
+
+
--- a/src/targets/java/unirest.js
+++ b/src/targets/java/unirest.js
@@ -1,0 +1,53 @@
+'use strict';
+
+var util = require('util');
+
+module.exports = function(options) {
+	var opts = util._extend({
+		indent : '			'
+	}, options);
+
+	var code = [];
+
+	code.push(util.format(
+			'HttpResponse<JsonNode> jsonResponse = Unirest.%s("%s")', this.source.method.toLowerCase(),
+			this.source.fullUrl));
+	
+	// construct headers
+	if (this.source.headers && this.source.headers.length) {
+		this.source.headers.map(function(header) {
+			code.push(util.format('.header("%s", "%s")', header.name,
+					header.value));
+		});
+	}
+	
+
+	//construct cookies argument
+	if (this.source.cookies && this.source.cookies.length) {
+		var cookies = this.source.cookies.map(function(cookie) {
+			return encodeURIComponent(cookie.name) + '='
+					+ encodeURIComponent(cookie.value);
+		});
+		code.push(util.format('.header("Cookie", "%s")', cookies.join('; ')));
+	}
+
+	//construct postdata
+	if (this.source.postData) {
+		if(this.source.postData.text){
+			code.push(util.format('.body(%s)', JSON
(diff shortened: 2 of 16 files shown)


src/targets/java/index.js
```diff
@@ -0,0 +1,5 @@
+'use strict';
+
+module.exports = require('require-directory')(module);
+
+
```
ahmadnassri, line 5: you should use the `.editeconfig` file to make sure your files are properly spaced....

src/targets/java/unirest.js
```diff
@@ -0,0 +1,53 @@
...
+
+module.exports = function(options) {
+	var opts = util._extend({
+		indent : '			'
+	}, options);
+
+	var code = [];
+
+	code.push(util.format(
+			'HttpResponse<JsonNode> jsonResponse = Unirest.%s("%s")', this.source.method.toLowerCase(),
```
ahmadnassri: please don't use `Unirest.METHOD`, it needs to support ANY method name, not just the Unirest predefined methods

src/targets/java/unirest.js
```diff
@@ -0,0 +1,53 @@
...
+			this.source.fullUrl));
+	
+	// construct headers
+	if (this.source.headers && this.source.headers.length) {
```
ahmadnassri: you should just use `source.allHeader` that way you don't have to do cookies below.

src/targets/java/unirest.js
```diff
@@ -0,0 +1,53 @@
...
+		this.source.headers.map(function(header) {
+			code.push(util.format('.header("%s", "%s")', header.name,
+					header.value));
+		});
+	}
+	
+
+	//construct cookies argument
+	if (this.source.cookies && this.source.cookies.length) {
```
ahmadnassri: see note above.

src/targets/java/unirest.js
```diff
@@ -0,0 +1,53 @@
...
+
+	//construct postdata
+	if (this.source.postData) {
+		if(this.source.postData.text){
+			code.push(util.format('.body(%s)', JSON
+					.stringify(this.source.postData.text)));
+		}
+	}
+
+	code.push(".asJson();");
```
ahmadnassri: this seems wrong, you're not checking for the postData.mimeType so how do you know its always JSON?

test/tests/java/unirest.js
```diff
@@ -0,0 +1,6 @@
+'use strict';
+
+require('should');
+
+module.exports = function (snippet, fixtures) {
```
ahmadnassri, line 5: you should add a test for the `indent` param here.

src/targets/java/unirest.js
```diff
@@ -0,0 +1,53 @@
...
+			return encodeURIComponent(cookie.name) + '='
+					+ encodeURIComponent(cookie.value);
+		});
+		code.push(util.format('.header("Cookie", "%s")', cookies.join('; ')));
+	}
+
+	//construct postdata
+	if (this.source.postData) {
+		if(this.source.postData.text){
+			code.push(util.format('.body(%s)', JSON
```
ahmadnassri: this seems wrong, you're not checking for `postData.mimeType`?

please review the [Guideline](https://github.com/ahmadnassri/httpsnippet/blob/master/CONTRIBUTING.md#creating-new-conversion-targets) for all the proper processing steps.

PR #23, opened 2015-03-11
OCaml updates

SGrondin:
OCaml updates


src/targets/ocaml/cohttp.js
```diff
@@ -7,9 +7,11 @@ module.exports = function (source, options) {
     indent: '  '
   }, options)
 
+  var methods = ['get', 'post', 'head', 'delete', 'patch', 'put', 'options']
```
ahmadnassri: wait, why are you forcing methods?! can't your library accept abstract methods? you really shouldn't be forcing a limit.

SGrondin:
merrrrgeeeeeeee

PR #24, opened 2015-03-12
Add Python target

montanaflynn:
Using Python 3's [native http client](https://docs.python.org/3/library/http.client.html). 

:snake:

### DIFF
--- a/src/targets/python/python3.js
+++ b/src/targets/python/python3.js
@@ -10,15 +10,15 @@ module.exports = function (source, options) {
   // Check which protocol to be used for the client connection
   var protocol = source.uriObj.protocol
   if (protocol === 'https:') {
-    code.push(util.format('conn = http.client.HTTPSConnection("%s")', source.uriObj.host))
+    code.push(util.format('conn = http.client.HTTPSConnection("%s")\n', source.uriObj.host))
   } else {
-    code.push(util.format('conn = http.client.HTTPConnection("%s")', source.uriObj.host))
+    code.push(util.format('conn = http.client.HTTPConnection("%s")\n', source.uriObj.host))
   }
 
   // Create payload string if it exists
   var payload = JSON.stringify(source.postData.text)
   if (payload) {
-    code.push(util.format('payload = %s', payload))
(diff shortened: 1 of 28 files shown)


src/targets/python/python3.js
```diff
@@ -0,0 +1,75 @@
...
+    code.push(util.format('conn.request("%s", "%s")', method, path))
+  }
+
+  // Get Response
+  code.push('\nres = conn.getresponse()')
+  code.push('data = res.read()')
+  code.push('\nprint(res.status)')
+  code.push('print(data)')
+
+  // console.log(code)
```
ahmadnassri, line 66: ^

test/fixtures/output/python/python3/application-json.py
```diff
@@ -0,0 +1,15 @@
+import http.client
+
+conn = http.client.HTTPConnection("mockbin.com")
+
+payload = "{\"number\": 1, \"string\": \"f\\\"oo\", \"arr\": [1, 2, 3], \"nested\": {\"a\": \"b\"}, \"arr_mix\": [1, \"a\", {\"arr_mix_nested\": {}}] }"
```
ahmadnassri, line 5: last param does not look escaped.

montanaflynn, line 5: It's the exact same as all the other fixtures, see [node for example](https://github.com/ahmadnassri/httpsnippet/blob/master/test/fixtures/output/node/native/application-json.js#L26)

Anyways I'm running the following to escape the payload like the other targets, maybe this should be abstracted away as a helper.

```
JSON.stringify(source.postData.text)
```

src/targets/python/python3.js
montanaflynn, line 66: Remove the commented out console.log?

test/fixtures/output/python/python3/application-json.py
ahmadnassri, line 5: ah, the syntax highlighting in github threw me off .. was checking this on mobile ...

PR #26, opened 2015-03-12
added custom methods support

shashiranjan84:


src/targets/java/unirest.js
```diff
@@ -12,36 +12,35 @@ module.exports = function (options) {
...
-  code.push('.asString();')
-  return code.join('\n' + opts.indent)
+  code.push(opts.indent + '.asString();')
+  return code.join('\n')
 }
 
 module.exports.info = {
   key: 'unirest',
-  title: 'Unirest',
+  title: 'JAVA',
```
ahmadnassri: please rebase, I've updated this to proper values, you're overwriting them again.

test/fixtures/available-targets.json
```diff
@@ -21,6 +21,20 @@
     "extname": ".sh"
   },
   {
+    "key": "java",
```
ahmadnassri: same as above, this is already added, please update your repo.

.project
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
ahmadnassri: please don't commit this ...

PR #29, opened 2015-03-13
Use proper quoting in HTTPie shell commands

jkbrzt:
This looks like an interesting project :+1: I've noticed a problem with quoting of the shell commands and this pull request fixes that for HTTPie:

Before, the resultant shell command was completely unquoted. This change introduces proper quoting of values and the URL - when necessary.

 It uses single quotes which provide "[strong quoting](http://wiki.bash-hackers.org/syntax/quoting#strong_quoting)" (only nested single quote characters need to be taken care of)  and  also make the output more readable as double quotes, ubiquitous in JSON, don't require escaping.

Before:

``` bash
 echo "{\"foo\": \"bar\"}" |  \ 
  http POST http://mockbin.com/request?foo=bar&foo=baz \
  accept:application/json \
  content-type:application/json \
  cookie:foo=bar; bar=baz
```

After:

``` bash
  echo '{"foo": "bar"}' |  \                                  # Improved readability
    http POST 'http://mockbin.com/request?foo=bar&foo=baz' \  # Quoted "&"
    accept:application/json \                                 # No quoting necessary
    content-type:application/json \                           # No quoting necessary
    cookie:'foo=bar; bar=baz'                                 # Quoted ";"
```

---

The cURL command (and possible also Wget) most likely suffer from improper quoting as well. It's not completely missing, as was the case of HTTPie, but it uses double quotes so literals like `"foo bar $HOME baz"` might get expanded by the shell. I think the same quoting mechanism should be used there as well.

### DIFF
--- a/src/targets/httpie.js
+++ b/src/targets/httpie.js
@@ -2,6 +2,18 @@
 
 var util = require('util')
 
+var shellQuote = function (value) {
+  // Unless `value` is a simple shell-safe string, quote it.
+  var shellSafe = /^[a-z0-9-_/.@%^=:]+$/i
+  if (!shellSafe.test(value)) {
+    // Use "strong quoting" using single quotes so that we only need
+    // to deal with nested single quote characters.
+    // <http://wiki.bash-hackers.org/syntax/quoting#strong_quoting>
+    return util.format("'%s'", value.replace(/'/g, "'\\''"))
+  }
+  return value
+}
+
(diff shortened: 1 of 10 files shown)


thibaultcha:
:+1:

ahmadnassri:
nice :+1: thanks @jakubroztocil I will integrate this with curl and wget as well.

PR #40, opened 2015-03-18
Add Ruby target language

montanaflynn:


src/targets/ruby/native.js
```diff
@@ -0,0 +1,66 @@
+'use strict'
+
+var util = require('util')
+
+module.exports = function (source, options) {
+  var self = this
+  var code = []
+  code.push('require \'uri\'')
```
ahmadnassri, line 8: if you use double quotes when single quotes are used inline, the linter wont complain.

PR #57, opened 2015-03-25
C# module

shashiranjan84:


src/targets/c#/restsharp.js
```diff
@@ -0,0 +1,51 @@
...
+  } else {
+    code.push(util.format('var client = new RestClient("%s");', source.fullUrl))
+    code.push(util.format('var request = new RestRequest(Method.%s);', source.method.toUpperCase()))
+
+  }
+
+  // construct headers
+  if (source.headers.length) {
+    source.headers.forEach(function (header) {
+      code.push(util.format('request.AddHeader("%s", "%s");', header.name, header.value))
```
thibaultcha: This could just be:

```
code.push('request.AddHeader("%s", "%s");', header.name, header.value)
```

src/targets/c#/restsharp.js
```diff
@@ -0,0 +1,51 @@
+'use strict'
+
+var util = require('util')
+
+var CodeBuilder = require('../../helpers/code-builder')
+
+module.exports = function (source, options) {
+  var opts = util._extend({
+    indent: '  '
```
thibaultcha: I don't think you need this, nor the `opts` object at all, neither in the CodeBuilder() constructor, as none of the generated code is indented or has options.

And since no more need for utils.format, you can remove the util loading at the top of the file.

PR #58, opened 2015-03-26
C module

shashiranjan84:
Using libcurl library

### DIFF
--- a/src/targets/c/libcurl.js
+++ b/src/targets/c/libcurl.js
@@ -6,6 +6,8 @@ module.exports = function (source, options) {
   var code = new CodeBuilder()
 
   code.push('CURL *hnd = curl_easy_init();')
+  code.push('')
+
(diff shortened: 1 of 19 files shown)


ahmadnassri:
can you add some line spacing for better readability?

shashiranjan84:
spacing in js file or generated snippet?

ahmadnassri:
in generated output.

shashiranjan84:
done.

src/targets/c/libcurl.js
```diff
@@ -0,0 +1,49 @@
+'use strict'
+
+var CodeBuilder = require('../../helpers/code-builder')
+
+module.exports = function (source, options) {
+  var code = new CodeBuilder()
+
+  code.push('CURL *hnd = curl_easy_init();')
+  code.push('')
```
ahmadnassri, line 9: there's a helper called `blank()` that does this instead.

PR #60, opened 2015-03-29
Java implementation using OkHttp

shashiranjan84:


shashiranjan84:
one of the four job failed though trvis build worked without any failure in my local

shashiranjan84:
I notice some bug, will send another pull request.

src/targets/java/okhttp.js
```diff
@@ -0,0 +1,79 @@
...
+module.exports = function (source, options) {
+  var opts = util._extend({
+    indent: '  '
+  }, options)
+
+  var code = new CodeBuilder(opts.indent)
+
+  var methods = [ 'GET', 'POST', 'PUT', 'DELETE', 'PATCH', 'HEAD']
+
+  var methodsWithBody = [ 'POST', 'PUT', 'DELETE', 'PATCH']
```
ahmadnassri, line 25: isn't this redundant? all methods could technically have bodies! should just focus on whether the postData has content instrlead.

shashiranjan84, line 25: OkHttp does not support get and head with body plus other methods 'POST',
'PUT', 'DELETE', 'PATCH' need to have body or passed null to api.

http://square.github.io/okhttp/javadoc/index.html

PR #63, opened 2015-04-15
Add option to include boilerplate

montanaflynn:
Added an option so you can decide wether to include the boilerplate code such as the package name, imports and main function. Defaults to true so nothing will change for existing implementations. 

Will need to update the [wiki](https://github.com/Mashape/httpsnippet/wiki/Targets#go) with the new option.

### DIFF
--- a/src/targets/go/native.js
+++ b/src/targets/go/native.js
@@ -27,11 +27,13 @@ module.exports = function (source, options) {
 
   var errorPlaceholder = opts.checkErrors ? 'err' : '_'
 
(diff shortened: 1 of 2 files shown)


ahmadnassri:
maybe add a test first?

montanaflynn:
Sure, can you link to an example of testing an option for reference?

ahmadnassri:
https://github.com/Mashape/httpsnippet/blob/master/test/tests/shell/curl.js

PR #77, opened 2016-03-16
Add option for --data-binary flag in curl

RogueBurger:
There is currently no way to generate a `--data-binary` flag for curl. My PR has a simple proof of concept for implementing it, but there are probably other ways it could be done as well. Is this functionality you would be willing to include in the codebase?


ahmadnassri:
looks good to me :+1: thanks

ahmadnassri:
would be great if you can add a test for that too?

PR #78, opened 2016-03-16
Add test for previous PR

RogueBurger:
As requested, I added a test to provide coverage for the new curl option.


ahmadnassri:
:clap: thank you !

PR #80, opened 2016-07-27
Fixed demo code snippets.

Gregen11:
The demo code snippets were failing in Node due to a misplaced semicolon and declaring/using the wrong variable name.


ahmadnassri:
excellent, thanks!

PR #83, opened 2016-08-09
Removed toLowerCase for headers, keeping the specified casing 

### LINKED ISSUE #74: Don't convert header names to lowercase
This is a super minor issue of aesthetics, but since HTTP header names are case insensitive, it would be nice if the header names are preserved as provided in the HAR object. I just think it looks nicer in my documentation to have:

```
-H 'Authorization: Bearer <Your Token>'
```

Instead of:

```
-H 'authorization: Bearer <Your Token>'
```

RolphH:
Small PR to fix #74 

Keep preventing duplication of headers a responsibility of the user.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -71,7 +71,7 @@ HTTPSnippet.prototype.prepare = function (request) {
   if (request.headers && request.headers.length) {
     // loweCase header keys
     request.headersObj = request.headers.reduceRight(function (headers, header) {
-      headers[header.name.toLowerCase()] = header.value
+      headers[header.name] = header.value
       return headers
     }, {})
   }


ahmadnassri:
the purpose here is to remove any duplication. if you want to keep the header casing then the appropriate solution is to de-dup in a different way.

kennethklee:
also note, according to http spec, iirc, header names are case insensitive

RolphH:
They are case insensitive, and according to HTTP spec 1.1 duplicate headers may exists. So why remove any duplication?

In #74 it is mentioned that keeping the casing is more an issue of aesthetics.

ahmadnassri:
@RolphH you're correct on both points. I believe _(having briefly inspected the flow, a deeper dive might be needed)_ there is an issue here of controlling the [`content-type`](https://github.com/Mashape/httpsnippet/blob/master/src/index.js#L123) header by relying on `headersObj` ... 

it's certainly not the best approach, and can be improved, to satisfy all three conditions: 
1. allow duplication
2. allow case-insensitive 
3. uniform access to `content-type` and other useful headers

I don't have a solution in mind, but I will experiment later, if you want to have a go at it, please be my guest.

deepakpathania-zz:
@ahmadnassri I believe all the points discussed are valid here and the headers should not be lowercased just to control one statement. A basic check(can be generalized) can suffice in the current scenario with the removed `toLowerCase()`. 
Something like : 
```javascript
var contentType;
for(var i in request.headersObj) {
    if(i.toLowerCase() == 'content-type') {
    contentType = i;
    }
}
request.headersObj[contentType] =  'multipart/form-data; boundary=' + form.getBoundary()
```
This will ensure that the `headersObj` is populated with the same case as intended in the original request. Let me know if this sounds okay and I would be happy to submit a PR.

RolphH:
What is needed to have this fixed?

darrenjennings:
@RolphH I'm new to the project 👋 and currently going through issues...

Looks like we can move this forward by removing `content-type` dependency in `headersObj` while making headers reflect the case sensitivity passed in. This PR has been inactive for almost a year, won't satisfy the constraints and lacks tests, so I'm closing, while I plan on getting a fix in to satisfy https://github.com/Kong/httpsnippet/issues/74 as part of our current httpsnippet housecleaning thanks for contributing @RolphH

PR #89, opened 2016-12-05
Swift and ObjC cannot handle JSON Obj with null value

lott-ai:
How to reproduce

```
var snippet = new HttpSnippet({
  "url": "http://mockbin.com/har",
  "method": "POST",
  "headers": [{
    "name": "content-type",
    "value": "application/json"
  }],
  "postData": {
    "params": [],
    "text": "{\"foo\":null}",
    "stored": true,
    "mimeType": "application/json",
    "size": 0,
    "jsonObj": {
      "foo":null
    },
    "paramsObj": false
  }
});

snippet.convert("swift");
snippet.convert("objc");
```

Since "foo" is null you will receive "Uncaught TypeError: Cannot read property 'toString' of null";

Added a try-catch around https://github.com/Mashape/httpsnippet/blob/master/src/targets/swift/helpers.js#L77 and https://github.com/Mashape/httpsnippet/blob/master/src/targets/objc/helpers.js#L67.

Also added a test for swift & objc and output tests for the request.


ahmadnassri:
excellent, thanks for this!

src/targets/objc/helpers.js
```diff
@@ -64,7 +64,11 @@ module.exports = {
       case '[object Boolean]':
         return value ? '@YES' : '@NO'
       default:
-        return '@"' + value.toString().replace(/"/g, '\\"') + '"'
+        try {
+          return '@"' + value.toString().replace(/"/g, '\\"') + '"'
```
ahmadnassri: wouldn't this be better if it was a `null` check rather than a try/fail?

lott-ai: Yes of course ```if (value !== null && value !== undefined) {}``` will also work if you prefer

lott-ai: Its probably more efficient too, so I will change this

lott-ai:
@ahmadnassri any updates on this?

PR #92, opened 2017-03-14
Use multiple --data flags in Curl target

gschier:
Curl support multiple `--data` flags for `application/x-www-form-urlencoded`. Here's an example:

```shell
# Before
curl -X POST http://foo.com \
  --data 'foo=bar&baz=qux'

# After
curl -X POST http://foo.com \
  --data foo=bar \
  --data baz=qux
```

From the [Curl docs](https://curl.haxx.se/docs/manpage.html#-d):

> If any of these options is used more than once on the same command line, the data pieces specified will be merged together with a separating &-symbol. Thus, using '-d name=daniel -d skill=lousy' would generate a post chunk that looks like 'name=daniel&skill=lousy'.

This pull request adds a new condition to the Curl target to handle `application/x-www-form-urlencoded` and append the appropriate `--data` flags.

### DIFF
--- a/src/targets/shell/curl.js
+++ b/src/targets/shell/curl.js
@@ -54,6 +54,22 @@ module.exports = function (source, options) {
       })
       break
 
+    case 'application/x-www-form-urlencoded':
+      if (source.postData.params) {
+        source.postData.params.map(function (param) {
+          code.push(
+            '%s %s', opts.binary ? '--data-binary' : (opts.short ? '-d' : '--data'),
+            helpers.quote(util.format('%s=%s', param.name, param.value))
+          )
+        })
+      } else {
+        code.push(
+          '%s %s', opts.binary ? '--data-binary' : (opts.short ? '-d' : '--data'),
+          helpers.escape(helpers.quote(source.postData.text))
+        )
+      }
+      break
+
     default:
       // raw request body
       if (source.postData.text) {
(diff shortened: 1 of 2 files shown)


thibaultcha:
Nice one! :+1:

garnold:
@gschier :1st_place_medal:

ahmadnassri:
excellent! Thank you!

gschier:
Wow, you folks are so responsive. Thank you! Is there a timeline for updating the npm package?

ahmadnassri:
@gschier well, you made it easy 😉

I'll try to push out a release in the next couple of hours.

gschier:
Awesome 💯

src/targets/shell/curl.js
```diff
@@ -54,6 +54,22 @@ module.exports = function (source, options) {
       })
       break
 
+    case 'application/x-www-form-urlencoded':
+      if (source.postData.params) {
+        source.postData.params.map(function (param) {
```
ahmadnassri, line 59: `forEach` is probably better here

gschier: Yes, that would be more fitting. I just copy/pasta'd from 10 lines above.

ahmadnassri:
didn't end up releasing, did some needed refactoring for performance. will continue tomorrow

gschier:
@ahmadnassri thanks for the heads up. I'll keep an eye out.

PR #99, opened 2017-07-06
Add target for raw HTTP request string (RFC 7230)

irvinlim:
This target isn't really a programming language, but I found the need to build raw HTTP request strings from a HAR object. I'll just be sharing this in case the maintainers here feel it might be useful to merge in.

The output is a HTTP/1.1 request string which is fully compliant with [RFC 7230](https://tools.ietf.org/html/rfc7230)/[7231](https://tools.ietf.org/html/rfc7231), the up-to-date set of IETF standards for HTTP/1.1. The name of the target or folder can be changed if anyone has a better idea.

Since I didn't know if this will be merged in, I haven't written any tests or updated the documentation. If we decide to go ahead with it, then I will add tests and docs prior to this being merged in.


darrenjennings:
@irvinlim We are in a bit of a transition time, but this looks good to proceed if you want to start the work. I like that you linked to the RFC, but I wonder if it makes more sense to call it `http/1.1` or some readable name `http1`. We can always link to the RFC in the comments/description. Tests/linting/docs would be great!

irvinlim:
Hey @darrenjennings, glad to hear that Kong is still maintaining this repo! Renaming it to `http/1.1` sounds good. I will add tests, linting and docs once I have the time this weekend or so.

darrenjennings:
our test suite is now stable on `master` if you would kindly rebase and add output fixtures and fix and linting issues. See #128 for inspiration.

irvinlim:
Hope everything's fine now! I didn't have time to come up with additional tests, but feel free to add any tests for custom options and such.

Fixtures were initially failing because CRLFs weren't committed in Git, you'll have to disable Git from converting CRLF to LF:

```
git config core.autocrlf false
```

darrenjennings (approved):

darrenjennings:
Great work @irvinlim and thanks again for your contribution.

PR #119, opened 2018-11-28
Update package.json

cdifulvio:
Fixes flatmap-stream vulnerability fix:

https://github.com/dominictarr/event-stream/issues/116

### DIFF
--- a/package.json
+++ b/package.json
@@ -80,7 +80,7 @@
     "chalk": "^1.1.1",
     "commander": "^2.9.0",
     "debug": "^2.2.0",
-    "event-stream": "^3.3.2",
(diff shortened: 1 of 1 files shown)


em92:
it should be ```"event-stream": "3.3.4"``` not ```"event-stream": "^3.3.4"```

JaneX8:
Can we please fix this as soon as possible?

```
  Critical        Malicious Package

  Package         flatmap-stream

  Patched in      No patch available

  Dependency of   httpsnippet

  Path            httpsnippet > event-stream > flatmap-stream

  More info       https://nodesecurity.io/advisories/737
```

PR #138, opened 2019-04-29
fix(target/python) query string fix

karol-maciaszek:
This change fixes handling of query string parameters for python/request target.

Query params can be supplied both by `queryString` and `url` (http://www.softwareishard.com/blog/har-12-spec/#request). Both scenarios are supported here https://github.com/Kong/httpsnippet/blob/master/src/index.js#L164.

However `python/requests` was looking only into `queryString` which was a cause of bug.

### DIFF
--- a/test/targets/python/requests.js
+++ b/test/targets/python/requests.js
@@ -1,10 +1,12 @@
+/* global it */
+
 'use strict'
 
 require('should')
 
 module.exports = function (HTTPSnippet) {
   it('should support query parameters provided in HAR\'s url', function () {
-    var result = new HTTPSnippet({ "method": "GET", "url": "http://mockbin.com/har?param=value" }).convert('python', 'requests', {
+    var result = new HTTPSnippet({ 'method': 'GET', 'url': 'http://mockbin.com/har?param=value' }).convert('python', 'requests', {
       showBoilerplate: false
     })
 
--- a/test/targets/python/requests.js
+++ b/test/targets/python/requests.js
@@ -2,5 +2,13 @@
 
 require('should')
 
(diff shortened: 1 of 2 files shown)


darrenjennings:
Can you add a test that demonstrates the fixed behavior?

karol-maciaszek:
Sure, just added a test. Also updated description :)

test/targets/python/requests.js
```diff
@@ -1,6 +1,16 @@
...
 
-module.exports = function (snippet, fixtures) {
+module.exports = function (HTTPSnippet) {
+  it('should support query parameters provided in HAR\'s url', function () {
+    var result = new HTTPSnippet({ 'method': 'GET', 'url': 'http://mockbin.com/har?param=value' }).convert('python', 'requests', {
+      showBoilerplate: false
+    })
+
+    result.should.be.a.String()
+    result.should.eql('import requests\n\nurl = "http://mockbin.com/har"\n\nquerystring = {"param":"value"}\n\nresponse = requests.request("GET", url, params=querystring)\n\nprint(response.text)')
```
darrenjennings: looks good. only feedback is to make output human readable using backticks. I refactored the `go` tests recently you can use as an example. This helps with git diffs in a future PR if tests fail.

https://github.com/Kong/httpsnippet/blob/c8e0b09fa3b9db4a75c29de85470851b04745801/test/targets/go/native.js#L22

darrenjennings (changes requested):

test/targets/python/requests.js
karol-maciaszek: Done :)

darrenjennings (approved):

Issue #132, opened 2019-04-02
PHP Guzzle generator


4n70w4:
Very necessary!

https://github.com/guzzle/guzzle
http://docs.guzzlephp.org/en/stable/

PR #140, opened 2019-05-18
PHP Guzzle generator #132

robertoarruda:
PHP Guzzle generator #132


src/targets/php/guzzle6.js
```diff
@@ -0,0 +1,119 @@
...
+  if (!opts.noTags && opts.closingTag) {
+    code.blank()
+      .push('?>')
+  }
+
+  return code.join()
+}
+
+module.exports.info = {
+  key: 'guzzle6',
```
darrenjennings: It is not typical to specify a library version in the target name. Would remove, while you could keep it in the description. If we ever need to support other versions of guzzle, we can decide how best to support them, possibly via options if the syntax isn't too different

darrenjennings (changes requested):

darrenjennings:
If you are able, it would also be nice to install guzzle on our CI and add it to these files so that we can run the snippets in a real environment to ensure they are working (currently only doing this for a few targets):

https://github.com/Kong/httpsnippet/blob/master/test/fixtures/cli.json#L14
https://github.com/Kong/httpsnippet/blob/master/.travis.yml#L10

src/targets/php/guzzle6.js
robertoarruda: Okay, I agree too!
I'll do it

darrenjennings:
@robertoarruda looking good! though the php clients array in `cli.json` needs to be updated so that the request validations run for guzzle. It's been a while since I've done PHP but, looks like you might need a bootstrap script which invokes the fixtures, so it can run composer's vendor autoload.php file, in order that the `GuzzleHttp\Client` class will be available. It looks like the scripts are indeed passing if this is the case.

philsturgeon (approved):

erunion (approved):

reynolek:
@robertoarruda Hi! I saw @erunion approved this PR today and it drew my attention to it. Is this still something that you would like to get merged in? If so, would you mind resolving conflicts, and signing the above CLA and I can get it merged.

Thanks!

robertoarruda:
> @robertoarruda Hi! I saw @erunion approved this PR today and it drew my attention to it. Is this still something that you would like to get merged in? If so, would you mind resolving conflicts, and signing the above CLA and I can get it merged.
> 
> Thanks!

Sure! It's solved.

test/fixtures/output/php/guzzle/custom-method.php
```diff
@@ -0,0 +1,9 @@
+<?php
+
+$client = new \GuzzleHttp\Client();
+
+$response = $client->request('PROPFIND', 'http://mockbin.com/har', [
```
erunion: Can you update the target to only add the third param if it's got content? Should clean this example up a touch.

erunion (approved):
Super minor thing to clean up some of these snippets with no params but otherwise this still looks good!

test/fixtures/output/php/guzzle/custom-method.php
erunion: @robertoarruda Sorry, could you update the Guzzle target as well with this change? Tests are currently failing because they're generating snapshots that match the old style.

erunion:
@robertoarruda Sorry, could you update the Guzzle target as well with this change? Tests are currently failing because they're generating snapshots that match the old style.

dimitropoulos:
https://github.com/Kong/httpsnippet/pull/253

PR #142, opened 2019-05-27
Add csharp httpclient

ajdust:
Feedback and comments welcome. #141

Issue #141, opened 2019-05-27
C# HttpClient


ajdust:
Hi,

It looks like only RestSharp is here for C#. [HttpClient](https://docs.microsoft.com/en-us/dotnet/api/system.net.http.httpclient?view=netcore-2.2) is in the .NET standard library and is the standard way to make HTTP requests (without a third party library like RestSharp).

If there is no ongoing work to add HttpClient, I can PR to add a snippet for HttpClient.

Thanks.

darrenjennings:
Would gladly welcome this, thanks @johnsabr.

h4ck1th4rd:
Any news for adding C# HttpClient?, Thank you.

PR #149, opened 2019-10-05
Add C# HttpClient target client

Atvaark:
Adds a second C# target client, one that doesn't require an additional NuGet package to work.
This will generate a C# snippet which can be pasted into the body of an existing (async) method.
#141 

Differences between this and PR #142:
- this also generates working code for *multipart/form-data* requests
- this supports custom HTTP methods
- this enables automatic gzip, deflate (but no br) decompression if a compatible accept header is found
- this generates code similar to the *restsharp* target client: no method declaration boilerplate is generated and the user has to add the required using statements on their own

### DIFF
--- a/src/targets/csharp/httpclient.js
+++ b/src/targets/csharp/httpclient.js
@@ -2,24 +2,64 @@
 
 var CodeBuilder = require('../../helpers/code-builder')
 
+function getDecompressionMethods (source) {
+  var acceptEncoding = source.allHeaders['accept-encoding']
+  if (!acceptEncoding) {
+    return [] // no decompression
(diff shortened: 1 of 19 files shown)


PavelStefanov:
Hey guys! This PR looks nice and really useful, can we merge that? If you need any help feel free to ask me, I'd be happy to help.
I'd suggest make HttpClient the default target

Atvaark:
Would anyone mind taking a look at this? The CI and tests work again and there aren't any conflicts.
@reynolek @darrenjennings

reynolek (approved):

PR #153, opened 2019-11-08
feat: change the default response code for Python Requests

dylanirlbeck:
When printing the response using the Python Requests client, the default is `print(response.text)`. From what I've seen, most developers use `print(response.json())` or `print response.json()` to print the output of a response. Thus, it makes more sense to have the default print value be in JSON format, rather than raw text format.

### DIFF
--- a/src/targets/python/requests.js
+++ b/src/targets/python/requests.js
@@ -89,7 +89,7 @@ module.exports = function (source, options) {
       .blank()
 
       // Print response
-      .push('print(response.text)')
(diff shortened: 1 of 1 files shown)


dylanirlbeck:
I can go ahead and change the testing data to get the build to pass if the maintainers think this is a beneficial change. Just let me know!

dimitropoulos:
closed by https://github.com/Kong/httpsnippet/pull/181

PR #174, opened 2020-08-07
feat: updating javascript targets to use `const` instead of `var`

erunion:
* Updates all Javascript and Node targets to use `const` instead of `var`.
* Fixes some inconsistencies with the `unirest` target where it was adding unnecessary linebreaks and indents in snippets.
* In the `jquery` target, I've added a semicolon to the `settings` object to stay consistent with the rest of the snippet.

### DIFF
--- a/src/targets/node/request.js
+++ b/src/targets/node/request.js
@@ -22,7 +22,7 @@ module.exports = function (source, options) {
   var includeFS = false
   var code = new CodeBuilder(opts.indent)
 
-  code.push('var request = require("request");')
+  code.push("const request = require('request');")
       .blank()
 
   var reqOpts = {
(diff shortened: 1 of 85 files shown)


src/targets/node/request.js
```diff
@@ -90,21 +90,21 @@ module.exports = function (source, options) {
...
     source.cookies.forEach(function (cookie) {
-      code.push('jar.setCookie(request.cookie("%s=%s"), "%s");', encodeURIComponent(cookie.name), encodeURIComponent(cookie.value), url)
+      code.push("jar.setCookie(request.cookie('%s=%s'), '%s');", encodeURIComponent(cookie.name), encodeURIComponent(cookie.value), url)
     })
     code.blank()
   }
 
   if (includeFS) {
-    code.unshift('var fs = require("fs");')
+    code.unshift("const fs = require('fs');")
```
reynolek, line 104: There are a few instances where you are changing the quotes from single to double in this file. However its not consistent. Please switch these to all single quotes to maintain consistency with the rest of the code.

reynolek (changes requested):

src/targets/node/request.js
erunion: @reynolek You mean in the generated code, or this target itself?

reynolek: Ugh thats a fair question ... the 2 places you did this do in fact take into account making the generated code use single quotes, and you wrapped them in doubles.

This is probably OK then, as I dont really want to see `\'` in random places either.

reynolek (approved):

PR #175, opened 2020-08-07
feat: updating php targets to use `[]` instead of `array()`

erunion:
Does what it says on the tin. Updates all PHP targets to use `[]` for arrays instead of `array()`. Also updates a few instances of `NULL` to the more consistent `null`.

### DIFF
--- a/src/targets/php/curl.js
+++ b/src/targets/php/curl.js
@@ -72,7 +72,7 @@ module.exports = function (source, options) {
     value: source.postData ? source.postData.text : undefined
   }]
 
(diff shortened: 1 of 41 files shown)


reynolek:
A benefit I can think of for keeping it to be `array()` is that it maintains support for php5 and `array()` still works in php7...That being said php5 is EOL as of Jan 2019.

erunion:
I thought the same, but yeah PHP 5 isn't getting any new updates now so I don't think there's any reason to code against that expectation.

PR #178, opened 2020-09-16
fix: don't convert header names to lowercase

### LINKED ISSUE #74: Don't convert header names to lowercase

jgiovaresco:
I've completed @darrenjennings branch to remove the `.toLowerCase` for header names.

Closes #74

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -73,9 +73,14 @@ HTTPSnippet.prototype.prepare = function (request) {
 
   // construct headers objects
   if (request.headers && request.headers.length) {
-    // loweCase header keys
+    var http2VersionRegex = /^HTTP\/2/
     request.headersObj = request.headers.reduce(function (headers, header) {
-      headers[header.name.toLowerCase()] = header.value
+      var headerName = header.name
+      if (request.httpVersion.match(http2VersionRegex)) {
+        headerName = headerName.toLowerCase()
+      }
+
+      headers[headerName] = header.value
       return headers
     }, {})
   }
--- a/test/index.js
+++ b/test/index.js
@@ -166,6 +166,24 @@ describe('HTTPSnippet', function () {
     done()
   })
 
+  it('should add "headersObj" to source object case insensitive', function (done) {
+    var fixture = Object.assign({}, fixtures.requests.headers)
+    fixture.headers = fixture.headers.concat({
+      name: 'Kong-Admin-Token',
+      value: 'Hunter1'
+    })
+
+    var req = new HTTPSnippet(fixture).requests[0]
+    req.headersObj.should.be.an.Object()
+    req.headersObj.should.eql({
+      'Kong-Admin-Token': 'Hunter1',
+      'accept': 'application/json',
+      'x-foo': 'Bar',
+    })
+
+    done()
+  })
+
   it('should modify orignal url to strip query string', function (done) {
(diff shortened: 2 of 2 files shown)


src/index.js
```diff
@@ -75,7 +75,7 @@ HTTPSnippet.prototype.prepare = function (request) {
   if (request.headers && request.headers.length) {
     // loweCase header keys
     request.headersObj = request.headers.reduce(function (headers, header) {
-      headers[header.name.toLowerCase()] = header.value
+      headers[header.name] = header.value
```
develohpanda: There is a comment in the linked issue mentioning that HTTP/2 _requires_ headers to be lowercase, while they are case-insensitive in HTTP/1.1 - https://github.com/Kong/httpsnippet/issues/74#issuecomment-561412157.

That's a great point, and since we have access to the version here (`request.httpVersion`) I am inclined suggest only converting to lowercase with HTTP/2, and use `header.name` as-is with other cases.

develohpanda (commented):
Thanks for finishing this off!

Do you have any thoughts about whether handling headers with HTTP/1.1 (case-insensitive) vs HTTP/2 (case-sensitive) is the responsibility of this library, or responsibility of the consumer? (cc: @darrenjennings)

I am impartial, but since we have access to the version on the HAR object, we might as well follow the [RFC](https://tools.ietf.org/html/rfc7540#section-8.1.2).

src/index.js
darrenjennings: Makes sense to me. @jgiovaresco can you add `httpVersion` detection and only lowercase on http/2.x?

re:
> Malformed Requests and Responses
A malformed request or response is one that is an otherwise valid
sequence of HTTP/2 frames but is invalid due to the presence of
extraneous frames, prohibited header fields, the absence of mandatory
header fields, or the inclusion of uppercase header field names.

src/index.js
```diff
@@ -75,7 +75,12 @@ HTTPSnippet.prototype.prepare = function (request) {
   if (request.headers && request.headers.length) {
     // loweCase header keys
     request.headersObj = request.headers.reduce(function (headers, header) {
-      headers[header.name.toLowerCase()] = header.value
+      var headerName = header.name
+      if (request.httpVersion.match(/^HTTP\/2/)) {
```
develohpanda: As a performance bump, it would be good to declare this regex once and re-use it rather than constructing it within each iteration of `.reduce()` 😄

develohpanda (changes requested):
Looks good now, nice to see an additional unit test! 👍

Just a minor comment on the use of regex

develohpanda (approved):
Nicely done 🙌

reynolek:
released in 1.23.0

erunion:
This change has unfortunately introduced some header duping in some cases of commonly used headers like `Content-Type` where if it's as `Content-Type` in the HAR, but a target is looking for `content-type` or adds its own `content-type` header (like in `multipart/form-data` where it is overridden) we end up with double headers containing potentially differing values.

I'm working on a PR right now though.

PR #154, opened 2019-11-13
Adding node-fetch target

hirenoble:
Hi Team,

As npm `request` package is no longer supported, it is essential to have code samples generated using `node-fetch`. I have attempted to write a target using `node-fetch`.

PR #179, opened 2020-09-18
Encode to JSON body for js-fetch target

jgiovaresco:
Closes #158

PR #180, opened 2020-09-29
Adding node-fetch target

### LINKED ISSUE #152: Node request is no longer supported
Hi Kong,

Recently there was an announcement that `request` will no longer be supported. It would be great to add on option using node `fetch`.

jgiovaresco:
@hirenoble is the original author (#154).
I've just resolve conflicts and fix tests that were failing.

@develohpanda noticed that JSON body should be stringified (https://github.com/node-fetch/node-fetch#post-with-json)
And looking at fetch options (https://github.com/node-fetch/node-fetch#fetch-options) I noticed that `json` attribute does not exist, so I've removed it.

Closes #152

### DIFF
--- a/src/targets/node/fetch.js
+++ b/src/targets/node/fetch.js
@@ -1,0 +1,111 @@
+/**
+ * @description
+ * HTTP code snippet generator for Node.js using node-fetch.
+ *
+ * @author
+ * @hirenoble
+ *
+ * for any questions or issues regarding the generated code snippet, please open an issue mentioning the author.
+ */
+
+'use strict'
+
+var stringifyObject = require('stringify-object')
+var CodeBuilder = require('../../helpers/code-builder')
+
+module.exports = function (source, options) {
+  var opts = Object.assign({
+    indent: '  '
+  }, options)
+
+  var includeFS = false
+  var code = new CodeBuilder(opts.indent)
+
+  code.push('const fetch = require(\'node-fetch\');')
+  var url = source.url
+  var reqOpts = {
+    method: source.method
+  }
+
+  if (Object.keys(source.queryObj).length) {
+    reqOpts.qs = source.queryObj
+  }
+
+  if (Object.keys(source.headersObj).length) {
+    reqOpts.headers = source.headersObj
+  }
+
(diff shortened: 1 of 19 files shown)


src/targets/node/fetch.js
```diff
@@ -0,0 +1,112 @@
...
+
+  switch (source.postData.mimeType) {
+    case 'application/x-www-form-urlencoded':
+      reqOpts.body = source.postData.paramsObj
+      break
+
+    case 'application/json':
+      if (source.postData.jsonObj) {
+        reqOpts.body = source.postData.jsonObj
+        reqOpts.json = true
```
develohpanda: Interesting, does this need to also be encoded to JSON body, or does the `reqOpts.json = true` handle it? I couldn't find docs on `node-fetch` about the json option does, but looking at the example, it looks as though we _do_ need to encode: https://github.com/node-fetch/node-fetch#post-with-json.

test/fixtures/output/node/fetch/application-json.js
```diff
@@ -0,0 +1,22 @@
+const fetch = require('node-fetch');
+
+let url = 'http://mockbin.com/har';
+
+let options = {
+  method: 'POST',
+  headers: {'content-type': 'application/json'},
+  body: {
+    number: 1,
+    string: 'f"oo',
```
develohpanda: `'f"oo'` 🤣

develohpanda (commented):
Thanks for sorting this PR out @jgiovaresco! Looks good to me. One comment regarding JSON encoding, similar to #179. Some of the fixtures would change after that as well.

src/targets/node/fetch.js
jgiovaresco: Good catch. Honestly I didn't take much attention on the existing code...

develohpanda: I hadn't reviewed it previously, so took the chance to now. Rest looks good though I think!

develohpanda (approved):
Nicely done 👏

hirenoble:
@jgiovaresco Thanks a lot for taking care of that! I wasn't noticing any activity on this thread.

PR #181, opened 2020-10-01
feat: change the default response code for Python Requests

### LINKED ISSUE #167: Using the 'json()' method in Python code that uses requests.
I use a program called [Stoplight Studio](https://github.com/stoplightio/studio) that in turn uses `httpsnippet` to generate sample code. I had a suggestion for the Python+requests code which I filed an issue with here: [#423](https://github.com/stoplightio/studio/issues/423#issuecomment-638121769).

Repeated here:
This is really minor. In the generated docs, under "Try It" -> "Code Generation" -> "Python/Requests", the last line of the code generated is:

```python
print(response.text)
```

The requests package has a `.json()` method, described here. The docs can check if the return data is `application/json`, and if so, suggest this line:

```python
print(response.json())
```

I think the question is whether `httpsnippet` is informed of the API response type.

Thanks!

cc @philsturgeon

jgiovaresco:
When printing the response using the Python Requests client, the default is `print(response.text)`. From what I've seen, most developers use `print(response.json())` or `print response.json()` to print the output of a response. Thus, it makes more sense to have the default print value be in JSON format, rather than raw text format.

@dylanirlbeck is the original author (#153).
I've just resolve conflicts and fix tests that were failing.

Closes #167

changelog(Improvements): Changed the default response code for Python Requests

### DIFF
--- a/src/targets/python/requests/client.ts
+++ b/src/targets/python/requests/client.ts
@@ -182,7 +182,7 @@ export const requests: Client<RequestsOptions> = {
     blank();
 
     // Print response
-    push('print(response.text)');
+    push('print(response.json())');
 
     return join();
   },
--- a/src/targets/python/requests/fixtures/application-form-encoded.py
+++ b/src/targets/python/requests/fixtures/application-form-encoded.py
@@ -10,4 +10,4 @@ headers = {"content-type": "application/x-www-form-urlencoded"}
 
 response = requests.post(url, data=payload, headers=headers)
 
-print(response.text)
+print(response.json())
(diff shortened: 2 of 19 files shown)


jgiovaresco:
@develohpanda @darrenjennings can we configure the repository to be able to add reviewers on PR?
If we can't should I ping you in a comment that a PR is available to review?

develohpanda:
For the time being, ping only @reynolek and myself as we are primarily maintaining this project now. I usually keep up with my notifications so will see it, and we can look towards updating the permissions long term. Does that sound good?

jgiovaresco:
👌I will do that 😉

antazoey:
`print(json.dumps(response.json(), indent=2))` looks nicer

dimitropoulos (approved):

PR #182, opened 2020-10-01
Print explicit error when using an invalid target

### LINKED ISSUE #87: They should be an explicit error when using an invalid target from command line
I use the command line.

If I write : 

`httpsnippet -t curl test.har`

I get : 

```
✓ test.har > curl [default] :
false
```

Since curl is not a valid target, I would expect the program to write an error message such as : `curl` is not a valid target, here is the list of valid targets : ...

jgiovaresco:
Screenshot:
Screenshot 2020-10-01 at 21 52 00

Closes #87

### DIFF
--- a/bin/httpsnippet
+++ b/bin/httpsnippet
@@ -56,6 +56,11 @@ cmd.args.forEach(function (fileName) {
     })
 
     .then(function (output) {
+      if (!output) {
+        var targetNames = HTTPSnippet.availableTargets().map(function (t) { return t.key }).join(', ')
+        return console.error('%s %s is not a valid target. Valid targets: %s', chalk.red('✖'), chalk.red(cmd.target), chalk.cyan(targetNames))
+      }
+
       // print
       if (!cmd.output) {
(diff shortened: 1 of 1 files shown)


jgiovaresco:
I had to confess that I'm not proud of the result... Would you be ok if I make some changes to try to modernize the code? 

At least I would like
- use ES6 syntax at least (const, let, arrow...)
- upgrade to Node 8

If you think we should keep compatibility with old node, I can configure Babel to transpile the result 🤔

jgiovaresco:
@reynolek @develohpanda 🙏

develohpanda (approved):
Nicely done 💯 

> use ES6 syntax at least (const, let, arrow...)

That would surely be nice, can be tackled in small chunks.

> upgrade to Node 8

But what is the actual change required here? `package.json` lists [node >= 4](https://github.com/Kong/httpsnippet/blob/225d9dcdfbb1355cc40f3278dc61ae8df38bb790/package.json#L41), so people are free to develop or use whatever version. Also looks like the Dockerfile uses a node 8 base image.

As an aside, because Node 8 will/has been deprecated, I would suggest upgrading to node 10 at the very least, if we do need to do anything here.

PR #183, opened 2020-10-06
chore: node 10.x & ES6 syntax

jgiovaresco:
As promised, here a 1st PR related to improving the codebase. 

Besides, I was wondering we could introduce [conventionalcommits](https://www.conventionalcommits.org/en/v1.0.0/) for commit messages. Using this would help the implementation of automatic release using [semantic-release](https://github.com/semantic-release/semantic-release) tools: it will allow us to generate automatically the CHANGELOG of the release.

If you agree we can enforce the convention using [commitlint](https://github.com/conventional-changelog/commitlint)

@develohpanda @reynolek

### DIFF
--- a/src/helpers/code-builder.js
+++ b/src/helpers/code-builder.js
@@ -1,6 +1,6 @@
 'use strict'
 
-var util = require('util')
+const util = require('util')
 
 /**
  * Helper object to format and aggragate lines of code.
@@ -11,7 +11,7 @@ var util = require('util')
  * @param {string} indentation Desired indentation character for aggregated lines of code
  * @param {string} join Desired character to join each line of code
  */
-var CodeBuilder = function (indentation, join) {
+const CodeBuilder = function (indentation, join) {
   this.code = []
   this.indentation = indentation
   this.lineJoin = join || '\n'
@@ -37,8 +37,8 @@ var CodeBuilder = function (indentation, join) {
  *   // returns: 'console.log("\t\thello world")'
  */
 CodeBuilder.prototype.buildLine = function (indentationLevel, line) {
-  var lineIndentation = ''
-  var slice = 2
+  let lineIndentation = ''
+  let slice = 2
   if (Object.prototype.toString.call(indentationLevel) === '[object String]') {
     slice = 1
     line = indentationLevel
@@ -52,7 +52,7 @@ CodeBuilder.prototype.buildLine = function (indentationLevel, line) {
     indentationLevel--
   }
 
-  var format = Array.prototype.slice.call(arguments, slice, arguments.length)
+  const format = Array.prototype.slice.call(arguments, slice, arguments.length)
   format.unshift(lineIndentation + line)
 
   return util.format.apply(this, format)
--- a/.jshintrc
+++ b/.jshintrc
@@ -1,5 +1,0 @@
-{
-  "asi": true,
-  "browser": true,
-  "node": true
-}
--- a/.npmignore
+++ b/.npmignore
@@ -1,3 +1,2 @@
-.jshintrc
 .editorconfig
 test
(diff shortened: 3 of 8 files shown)


src/helpers/code-builder.js
```diff
@@ -1,6 +1,6 @@
 'use strict'
 
-var util = require('util')
+const util = require('util')
```
develohpanda, line 3: Would be great to enforce this with ES Lint (now or eventually)

package.json
```diff
@@ -38,7 +38,7 @@
     "xmlhttprequest"
   ],
   "engines": {
-    "node": ">=4"
+    "node": ">=10"
```
develohpanda: I'm not familiar enough with the `engines` entry in `package.json`, but my understanding is that it will require consumers to use Node 10 and above, otherwise they will receive warnings?

Since we are not using anything specifically from Node 10, I'm not sure if we should change this yet. Changing travis and the dockerfile to use Node 10 seems fine, though.

develohpanda (commented):
Small comment, also apologies for the merge conflict from another PR 😄 

Also not opposed to introducing conventionalcommits + commitlint but I don't know for sure how it would fit in the workflow? Does that _require_ us to follow a particular PR merge strategy (merge vs squash etc)?

src/helpers/code-builder.js
erunion: @develohpanda I have a WIP PR that'll do that and can get that work a PRable state after this is accepted.

package.json
jgiovaresco: The engine will prevent users from using the library if they are using older Node. 
The only issue I see with not updating `engine` is that if we are using latest ES standard it would fail on older Node because it won't be supported. 
To ensure that we might configure Travis CI to run tests on older Node also?

develohpanda: As you suggested in a different PR, it might be good to configure Babel to transpile the result in that case.

While I agree it would be good to have consumers be on a newer version of Node, it feels unnecessary to make that a restriction when previously it wasn't, just so we can develop with the latest ES standard.

jgiovaresco:
@develohpanda I've reverted the change on the `engine` attribute. 

Concerning conventionalcommits + commitlint, I've always used it with a Rebase & Merge strategy. But I know that the tool ignores *merge* messages so it does not require anything related to the PR merge strategy

Do you agree to merge this PR or do you prefer I work on configuring Babel and update the source code after that?

develohpanda:
I'm guessing the main requirement of conventionalcommits is to keep commit messages in the main branch. I tend to merge PRs with a Squash & Merge strategy so that the PR can have any number of commits but the history in the default branch will remain clean. I'm not opposed to using a different strategy in this repository, though, so can switch to rebase and merge if that's what is required.

> I've reverted the change on the engine attribute.

I'm not totally familiar with upgrading older libraries like this one. My only concern is, even though older versions of Node are being deprecated, there are bound to be projects that still use them for whatever reason, and it would be good to make sure changes we make don't cause breaking issues for others, unless we have a valid reason to do so (in which case we do a major version bump). But I also agree that it would be nice to modernize the code-base so _we_ can work with it. So long as we can upgrade without significant impact to consumers, I'd be comfortable with that.

With that, we can potentially look towards Babel, or just look straight to TypeScript (given TS and JS can co-exist).

Let's work through these topics one at a time and with a plan - would you like to open separate issues with proposals for each of the following topics you have raised, so that they can be addressed and discussed before forging ahead with code?
- conventionalcommits + commitlint
- ESLint
- CD pipeline / automated deployments
- upgrading Node & ES6 + Babel (or TypeScript 👀)
- and anything else

> Do you agree to merge this PR or do you prefer I work on configuring Babel and update the source code after that?

Both options work since this library is never on the critical path / requires urgent releases. Given there are very few changes (only moving to `const` and `let`), I'd favor closing this and working through configuring Babel or TS first. 

How does that sound? 😄

PR #184, opened 2020-10-06
fix: can't load the form-data polyfill in browsers because fs doesn't exist

erunion:
This fixes a bug in the recent `multipart/form-data` work I did in #173 where instead of detecting if the loaded `FormData` object from the `form-data` module is a polyfill, we were loading the polyfill directly from that library. Problem with this is that that [file](https://github.com/form-data/form-data/blob/master/lib/form_data.js) loads the `fs` module, which doesn't exist in the browser.

Instead, we're now just relying on the `form-data` module to give us the appropriate export (for browsers it'll load [`browser.js`](https://github.com/form-data/form-data/blob/master/lib/browser.js)) it has given the environment HTTPSnippet is being run under.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -5,7 +5,6 @@
 var debug = require('debug')('httpsnippet')
 var es = require('event-stream')
 var MultiPartForm = require('form-data')
-var FormDataPolyfill = require('form-data/lib/form_data')
 var qs = require('querystring')
 var reducer = require('./helpers/reducer')
 var targets = require('./targets')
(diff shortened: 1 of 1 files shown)


develohpanda (approved):
👏

marcelltoth:
When can we expect a release including this fix?

develohpanda:
Hi @marcelltoth, likely late November or early December. Unfortunately the team is quite bogged down at the moment.

marcelltoth:
> 
> 
> Hi @marcelltoth, likely late November or early December. Unfortunately the team is quite bogged down at the moment.

That's sad to hear given that this is a complete blocker for any browser usage (it fails the build) and the fix is already merged.

But anyway, thanks @develohpanda for the quick reply, appreciated!

develohpanda:
Sorry about that! I'll raise it tomorrow to see if we can push out a release sooner. This is a project my team inherited, and I haven't had a chance to automate the deployments yet. 😢

erunion:
@marcelltoth If you're in a pinch for time you can use our fork that has the fix until it's published here. https://www.npmjs.com/package/@readme/httpsnippet

develohpanda:
Thanks @erunion! [1.24.0](https://github.com/Kong/httpsnippet/releases/tag/1.24.0) has been published so you should be good to go, @marcelltoth!

marcelltoth:
Thank you @develohpanda, you rock! 🎉

PR #187, opened 2020-10-15
fix: updating the node-fetch target to handle form-urlencoded requests

erunion:
The `node-fetch` target was originally sending `x-www-form-urlencoded` as a JSON object to `body` when it should be encoded with `URLSearchParams` as per the [docs](https://www.npmjs.com/package/node-fetch#post-with-form-parameters).

I also added a line break in the `multipart/form-data` handling so the `formData.append();` calls have a little bit of room to breathe.

### DIFF
--- a/src/targets/node/fetch.js
+++ b/src/targets/node/fetch.js
@@ -37,7 +37,15 @@ module.exports = function (source, options) {
 
   switch (source.postData.mimeType) {
     case 'application/x-www-form-urlencoded':
-      reqOpts.body = source.postData.paramsObj
+      code.unshift('const { URLSearchParams } = require(\'url\');')
+      code.push('const encodedParams = new URLSearchParams();')
+      code.blank()
+
+      source.postData.params.forEach(function (param) {
+        code.push('encodedParams.set(\'' + param.name + '\', \'' + param.value + '\');')
+      })
+
+      reqOpts.body = 'encodedParams.toString()'
       break
 
     case 'application/json':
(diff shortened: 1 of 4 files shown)


src/targets/node/fetch.js
```diff
@@ -37,7 +37,15 @@ module.exports = function (source, options) {
...
-      reqOpts.body = source.postData.paramsObj
+      code.unshift('const { URLSearchParams } = require(\'url\');')
+      code.push('const encodedParams = new URLSearchParams();')
+      code.blank()
+
+      source.postData.params.forEach(function (param) {
+        code.push('encodedParams.set(\'' + param.name + '\', \'' + param.value + '\');')
+      })
+
+      reqOpts.body = 'encodedParams.toString()'
```
DMarby: The `node-fetch` documentation and tests seems use pass the URLSearchParams object directly as the body, is there any particular reason for using `toString` instead?

DMarby (changes requested):
Thank you for this PR, looks great!
Just have one question regarding how the encodedParams are used.

src/targets/node/fetch.js
erunion: Seems that it [supports both URLSearchParams or a stringified body](https://github.github.io/fetch/#request-body), but I'll change it to just pass `encodedParams` directly.

erunion: Tested locally and  without `.toString()` (and with) it works fine:

```js
const { URLSearchParams } = require('url');
const encodedParams = new URLSearchParams();

encodedParams.set('foo', 'bar');

let url = 'http://httpbin.org/post';

let options = {
  method: 'POST',
  qs: {foo: ['bar', 'baz'], baz: 'abc', key: 'value'},
  headers: {
    accept: 'application/json',
    'content-type': 'application/x-www-form-urlencoded',
    cookie: 'foo=bar; bar=baz; '
  },
  body: encodedParams
};

fetch(url, options)
  .then(res => res.json())

/*
{ args: {},
  data: '',
  files: {},
  form: { foo: 'bar' },
  headers:
   { Accept: 'application/json',
     'Accept-Encoding': 'gzip,deflate',
     'Content-Length': '7',
     'Content-Type': 'application/x-www-form-urlencoded',
     Cookie: 'foo=bar; bar=baz;',
     Host: 'httpbin.org',
     'User-Agent': 'node-fetch/1.0 (+https://github.com/bitinn/node-fetch)',
     'X-Amzn-Trace-Id': 'Root=1-5f90a62c-221fa1af7ec36f472ed54945' },
  json: null,
  origin: '98.37.58.48',
  url: 'http://httpbin.org/post' }
*/
```

erunion:
@DMarby I've removed `.toString()` so this should be good to go!

DMarby (approved):

develohpanda (approved):
LGTM, thank you for doing this!

PR #188, opened 2020-10-16
fix: regression in header case-insensitivity

erunion:
This fixes a regression that was introduced in the recent header case-insensitivity work (https://github.com/Kong/httpsnippet/pull/178) that is causing edge cases where, on certain targets, it's possible for a header to be:

* Duplicated and both instances having differing data. This was mostly apparent on `multipart/form-data` requests where one header would just be `multipart/form-data` while the other would be that plus the `boundary`.
* Some targets would completely fail to interpret the header, resulting in either an `undefined` header being set, being set in ways that it shouldn't be, or the target completely failing to function because a header wasn't being removed.

To resolve this I've:

* Added a new `headers` helper with a couple methods to traverse the now case-insensitive object of headers so that these targets that need special case handling for something like `content-type` can safely search for that in the available headers.
* Updated every target that was looking for case-sensitive headers to now use this new helper to do the same work it needs to do.

For testing all of this, I've updated the `multipart/form-data` test fixture to have its `content-type` header as `Content-Type` to force this case-insensitive lookup for targets that need it.

😰

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -7,6 +7,7 @@ var es = require('event-stream')
 var MultiPartForm = require('form-data')
 var qs = require('querystring')
 var reducer = require('./helpers/reducer')
+var helpers = require('./helpers/headers')
 var targets = require('./targets')
 var url = require('url')
 var validate = require('har-validator/lib/async')
@@ -168,17 +169,11 @@ HTTPSnippet.prototype.prepare = function (request) {
 
         // Since headers are case-sensitive we need to see if there's an existing `Content-Type` header that we can
         // override.
-        let foundContentType = false
-        Object.keys(request.headersObj).forEach(header => {
-          if (header.toLowerCase() === 'content-type') {
-            foundContentType = true
-            request.headersObj[header] = 'multipart/form-data; boundary=' + boundary
-          }
-        })
+        const contentTypeHeader = helpers.hasHeader(request.headersObj, 'content-type')
+          ? helpers.getHeaderName(request.headersObj, 'content-type')
+          : 'content-type'
 
-        if (!foundContentType) {
-          request.headersObj['content-type'] = 'multipart/form-data; boundary=' + boundary
-        }
+        request.headersObj[contentTypeHeader] = 'multipart/form-data; boundary=' + boundary
       }
       break
 
--- a/src/index.js
+++ b/src/index.js
@@ -165,7 +165,20 @@ HTTPSnippet.prototype.prepare = function (request) {
         }
 
         request.postData.boundary = boundary
-        request.headersObj['content-type'] = 'multipart/form-data; boundary=' + boundary
+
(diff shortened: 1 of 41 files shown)


DMarby (approved):
Looks great, thank you for finding this regression and fixing it!

src/index.js
```diff
@@ -165,7 +165,20 @@ HTTPSnippet.prototype.prepare = function (request) {
...
         request.postData.boundary = boundary
-        request.headersObj['content-type'] = 'multipart/form-data; boundary=' + boundary
+
+        // Since headers are case-sensitive we need to see if there's an existing `Content-Type` header that we can
+        // override.
+        let foundContentType = false
+        Object.keys(request.headersObj).forEach(header => {
+          if (header.toLowerCase() === 'content-type') {
+            foundContentType = true
+            request.headersObj[header] = 'multipart/form-data; boundary=' + boundary
```
jgiovaresco: I think I miss something I don't understand why we would not use the new helper functions you have added?

we would have something like

```
const contentTypeHeader = helpers.hasHeader('content-type') ? helpers.getHeaderName('content-type'): 'content-type';
request.headersObj[contentTypeHeader] = 'multipart/form-data; boundary=' + boundary
```

jgiovaresco (commented):
Nice catch 👏 

I'm wondering if we should also use the helper for 

- https://github.com/Kong/httpsnippet/blob/675df1ae8574b660e8d1ee86c9e14eb229b012b6/src/targets/c/libcurl.js#L31
- https://github.com/Kong/httpsnippet/blob/675df1ae8574b660e8d1ee86c9e14eb229b012b6/src/targets/node/fetch.js#L79-L82

src/index.js
erunion: yeah this was an oversight on my part. will update!

erunion:
@jgiovaresco Updated with your feedback.

> I'm wondering if we should also use the helper for
> 
> * https://github.com/Kong/httpsnippet/blob/675df1ae8574b660e8d1ee86c9e14eb229b012b6/src/targets/c/libcurl.js#L31
> * https://github.com/Kong/httpsnippet/blob/675df1ae8574b660e8d1ee86c9e14eb229b012b6/src/targets/node/fetch.js#L79-L82

I'd love to swap the `reqOpts.headers` object with an object that has some functionality like [caseless](https://www.npmjs.com/package/caseless) so we could eliminate the need for `allHeaders` and this new helper utility added, but that's a larger refactor that I think is out of scope for resolving this regression.

test/headers.js
```diff
@@ -0,0 +1,43 @@
+/* global describe, it */
```
develohpanda, line 1: Nice tests!

develohpanda (approved):
`caseless` would be cool, but I agree out of scope for this regression.

Thank you for taking the time to investigate and fix this! 🎉 Sorry about the delay, I've been quite bogged down for a while!

PR #189, opened 2020-10-16
feat: cleaner python requests json snippets

erunion:
This cleans up the Python `requests` target snippets to:

* [x] Ensure that all dict keys are wrapped in double quotes (most everything already was).
* [x] For `application/json` requests, we now build out Python-compatible dicts and pass those to `requests` via the `json=payload` argument that it provides.
  * With this work, `false`, `true`, and `null` are now being translated into the Python-proper literals of `False`, `True`, and `None`.

### DIFF
--- a/src/targets/python/helpers.js
+++ b/src/targets/python/helpers.js
@@ -6,7 +6,7 @@ var util = require('util')
  * Create an string of given length filled with blank spaces
  *
  * @param {number} length Length of the array to return
- * @return {string}
+ * @param {string} str String to pad out with
  */
 function buildString (length, str) {
   return Array.apply(null, new Array(length)).map(String.prototype.valueOf, str).join('')
--- a/src/targets/python/helpers.js
+++ b/src/targets/python/helpers.js
@@ -1,0 +1,79 @@
+'use strict'
+
+var util = require('util')
+
+/**
+ * Create an string of given length filled with blank spaces
+ *
+ * @param {number} length Length of the array to return
+ * @return {string}
+ */
+function buildString (length, str) {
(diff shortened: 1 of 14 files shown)


src/targets/python/helpers.js
```diff
@@ -0,0 +1,79 @@
+'use strict'
+
+var util = require('util')
+
+/**
+ * Create an string of given length filled with blank spaces
```
DMarby, line 6: I think this function documentation needs to be updated, as it's missing the second argument which is what is used to build the resulting string out of

DMarby (changes requested):
Thank you for this contribution!
Just one minor note about the function documentation on buildString

erunion:
@DMarby Updated that docblock, and since I copied that function from `src/targets/swift/helpers.js` I've also updated it there since it was missing as well.

Once this library gets some Babel compilation, these string padding methods could probably removed in favor of `String.prototype.padStart` and `String.prototype.padEnd`.

DMarby (approved):
LGTM :shipit:

develohpanda (approved):
Thanks for doing this! I think a conflict was introduced by your other PR I just merged - #188 😅

erunion:
@develohpanda I've fixed the merge conflict so this should be good to go.

develohpanda:
Hi @erunion, it looks like [one of the tests](https://travis-ci.org/github/Kong/httpsnippet/jobs/746357535#L1719) from the merge conflict is now failing 🙁 just need to fix up the expected double-quotes in that I think!

erunion:
@develohpanda Oops, sorry about that. Fixed now.

PR #191, opened 2020-10-22
fix: clj-http handling of literal null JSON bodies

szakharchenko:
`jsToEdn()` expects` jsType(null)` to return `'null'`, but `null` has no fields, so `jsType(null)` fails. Add a special case for `null` to `jsType()` to make things work as expected.


jgiovaresco (commented):
Seems legit.

I'm wondering if we can add or update a test case to ensure this 🤔

dimitropoulos (approved):

Issue #161, opened 2020-04-15
Swift/Obj-c snippet conversions crash when given empty/invalid form data


pimterry:
When converting requests with a `application/x-www-form-urlencoded` content type, if there are no successfully parsed parameters in the body, the Swift & Obj-C conversions both crash with "Cannot read property 'name' of undefined". That happens because these lines both assume that there's at least one parameter:

https://github.com/Kong/httpsnippet/blob/master/src/targets/swift/nsurlsession.js#L49

https://github.com/Kong/httpsnippet/blob/master/src/targets/objc/nsurlsession.js#L49

Most (all?) cases where you'll hit this require technically invalid bodies, e.g. sending empty or JSON bodies as form data, but unfortunately that's not uncommon. It would be better to not crash in this case, as in every other target.

dimitropoulos:
I'm almost certain this has been fixed now.  Today, thanks to TypeScript, this code has been updated to wrap with `if (postData.params) {` which will prevent calling `.name` on a param value.

If I'm mistaken, please show some repro steps and we'd be happy to take a look!

pimterry:
This hasn't been fixed. Repro:

* Download [this HAR file](https://gist.github.com/pimterry/076102da7c03f9e851a3b69613379934) into the root of your HTTP Snippet folder
* In Node, run:
    ```javascript
    > har = require('./empty-params.har.json')
    > HS = require('.')
    > hs = new HS.HTTPSnippet(har)
    > hs.convert('swift', 'nsurlsession')
    Uncaught TypeError: Cannot read property 'name' of undefined
        at convert (.../dist/targets/swift/nsurlsession/client.js:75:82)
        at .../dist/httpsnippet.js:238:74
        at Array.map (<anonymous>)
        at HTTPSnippet.convert (.../dist/httpsnippet.js:238:42)
    ```

(Note that this requires the fix in #294 because otherwise you currently can't require httpsnippet in node at all).

I'll open a PR to fix it in a sec.

PR #192, opened 2020-10-29
fix: prevent crash in Swift/Objc with checking length of input body post params

### LINKED ISSUE #196: undefined httpBody with empty input post body.
**To reproduce the bug:**
the latest version of the httpsnippet.

1) Save this JSON as `example-bad.json` file.
```JSON
{
  "method": "POST",
  "url": "http://mockbin.com/har?key=value",
  "httpVersion": "HTTP/1.1",
  "queryString": [
    {
      "name": "foo",
      "value": "bar"
    },
    {
      "name": "foo",
      "value": "baz"
    },
    {
      "name": "baz",
      "value": "abc"
    }
  ],
  "headers": [
    {
      "name": "accept",
      "value": "application/json"
    },
    {
      "name": "content-type",
      "value": "application/x-www-form-urlencoded"
    }
  ],
  "cookies":  [
    {
      "name": "foo",
      "value": "bar"
    },
    {
      "name": "bar",
      "value": "baz"
    }
  ],
  "postData": {
    "mimeType": "application/x-www-form-urlencoded",
    "params": [
    ]
  }
}
```

2) generate Swift snippet with this command:
`httpsnippet example-bad.json --target swift --client unirest --output .`

this should be output:
```SWIFT
import Foundation

let headers = [
  "cookie": "foo=bar; bar=baz",
  "accept": "application/json",
  "content-type": "application/x-www-form-urlencoded"
]

let request = NSMutableURLRequest(url: NSURL(string: "http://mockbin.com/har?foo=bar&foo=baz&baz=abc&key=value")! as URL,
                                        cachePolicy: .useProtocolCachePolicy,
                                    timeoutInterval: 10.0)
request.httpMethod = "POST"
request.allHTTPHeaderFields = headers
request.httpBody = postData as Data

let session = URLSession.shared
let dataTask = session.dataTask(with: request as URLRequest, completionHandler: { (data, response, error) -> Void in
  if (error != nil) {
    print(error)
  } else {
    let httpResponse = response as? HTTPURLResponse
    print(httpResponse)
  }
})

dataTask.resume()

```

**Problem:**
you can see `request.httpBody = postData as Data` without defining `postData` earlier.

**Expected output:**
1 ) defining `postData` earlier (with empty value base on input JSON file)
or
2) generating the snippet without the `request.httpBody`.

irajtaghlidi:
related to #161

now the generator checks the length of input body params to prevent crash because the first element doesn't exist. So it generates snippets without the body, just like other target languages.

Also, fix #196 with setting `req.hasBody = false` when we have an empty  post data params in input.

changelog(Fixes): Fixed an issue that caused a crash in Swift/Objc with checking length of input body post params

### DIFF
--- a/src/targets/objc/nsurlsession/client.ts
+++ b/src/targets/objc/nsurlsession/client.ts
@@ -59,15 +59,18 @@ export const nsurlsession: Client<NsurlsessionOptions> = {
             // The user can just add/remove lines adding/removing body parameters.
             blank();
 
+            const [head, ...tail] = postData.params;
             push(
-              `NSMutableData *postData = [[NSMutableData alloc] initWithData:[@"${postData.params[0].name}=${postData.params[0].value}" dataUsingEncoding:NSUTF8StringEncoding]];`,
+              `NSMutableData *postData = [[NSMutableData alloc] initWithData:[@"${head.name}=${head.value}" dataUsingEncoding:NSUTF8StringEncoding]];`,
             );
 
-            for (let i = 1, len = postData.params.length; i < len; i++) {
+            tail.forEach(({ name, value }) => {
               push(
-                `[postData appendData:[@"&${postData.params[i].name}=${postData.params[i].value}" dataUsingEncoding:NSUTF8StringEncoding]];`,
+                `[postData appendData:[@"&${name}=${value}" dataUsingEncoding:NSUTF8StringEncoding]];`,
               );
-            }
+            });
+          } else {
+            req.hasBody = false;
           }
           break;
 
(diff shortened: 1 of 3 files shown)


irajtaghlidi:
we can also check the length of the param attribute of the input request in switch case, more like `application/json` case in the next lines.

```swift
if (source.postData.text || source.postData.jsonObj || source.postData.params) {
    req.hasBody = true

    switch (source.postData.mimeType) {
      case 'application/x-www-form-urlencoded':
      if(source.postData.params.length) { //+
           ...
      }  //+
      else { //+
        req.hasBody = false //+
      } //+
```

dimitropoulos (approved):
thankfully, TypeScript had already caught at least one of these bugs with the undefined params.  

thanks!

Issue #193, opened 2020-10-29
Implicitly coercing risk with Swift generated snippet.


irajtaghlidi:
The Swift snippet generator print `error` and `httpResponse` with **implicitly coercing** risk.

Xcode: 12.0.1

![snippet](https://user-images.githubusercontent.com/785830/97632545-b82cb200-1a43-11eb-8e95-e3acdd46e9d7.png)

dimitropoulos:
fixed in https://github.com/Kong/httpsnippet/pull/195

Issue #100, opened 2017-07-17
Swift code generation is outdated


domingguss:
Enums are lowerCamelCased nowadays, and paramaters should be made explicit.
Try testing it in a XCode8 Playground context.

```swift
import Foundation

let headers = [
    "authorization": "Bearer eyJhbGciOiJSUzI1NiJ",
    "content-type": "application/json"
]

var request = NSMutableURLRequest(url: URL(string: "http://foo.bar/api/items")!, cachePolicy: .useProtocolCachePolicy, timeoutInterval: 10.0)
request.httpMethod = "GET"
request.allHTTPHeaderFields = headers

let session = URLSession.shared
let dataTask = session.dataTask(with: request as URLRequest, completionHandler: { (data, response, error) -> Void in
    if let e = error {
        print(e)
    } else {
        if let httpResponse = response as? HTTPURLResponse {
            print(httpResponse)
        }
    }
})

dataTask.resume()
```

darrenjennings:
@domingguss would you be up for submitting a pull request?

dimitropoulos:
looks like this has been fixed, comparing the snippets of today to this description: https://github.com/Kong/httpsnippet/blob/master/src/targets/swift/nsurlsession/fixtures/full.swift

PR #194, opened 2020-10-29
Fix implicitly coercing warning in Swift snippet generator

irajtaghlidi:
related to #193 and #100

More details about the solution: https://stevenpcurtis.medium.com/expression-implicitly-coerced-from-string-to-any-why-swift-why-190dd0a58c58


irajtaghlidi:
A new PR submitted with the correct commits (a branch from master, instead of from the last PR, to prevent merging commits from two different PR) here #195

Please correct my method about how we should submit two different PRs for two different bugs, but both are working with the same files.

PR #198, opened 2020-10-31
fix: cUrl target should encode x-www-form-urlencoded post data params

### LINKED ISSUE #197: cUrl target does not encode body for application/x-www-form-urlencoded request
cUrl target does not URL encode when the request is a POST request with form URL encoded parameters.

Given the following HAR request 
```
{
   "method":"POST",
   "url":"http://localhost:3000/url-encoded",
   "httpVersion":"HTTP/1.1",
   "cookies":[],
   "headers":[
      {
         "name":"Content-Type",
         "value":"application/x-www-form-urlencoded"
      }
   ],
   "queryString":[],
   "postData":{
      "mimeType":"application/x-www-form-urlencoded",
      "text":"%24filter=my%26value%20%3D%201",
      "params":[
         {
            "name":"$filter",
            "value":"my&value = 1"
         }
      ],
      "size":0,
      "jsonObj":false,
      "paramsObj":{
         "$filter":"my&value = 1"
      }
   },
   "headersSize":-1,
   "bodySize":-1,
   "settingEncodeUrl":true
}
```

The current cUrl snippet is 

```
curl --request POST \
  --url http://localhost:3000/url-encoded \
  --header 'Content-Type: application/x-www-form-urlencoded' \
  --data '$filter=my&value = 1'
```

To tell to cUrl to URL encode the data, we should use `--data-urlencode`. This will encode the value, but we should also encode the key. It can be done by prefixing the string to encode with `=`.
[source](https://ec.haxx.se/http/http-post#url-encoding)

See https://github.com/Kong/insomnia/issues/1958

jgiovaresco:
Closes #197

Changelog(Fixes): Fixed an issue on how `x-www-form-urlencoded` post data parameters are encoded

### DIFF
--- a/src/fixtures/runCustomFixtures.ts
+++ b/src/fixtures/runCustomFixtures.ts
@@ -1,4 +1,5 @@
-import { readFile } from 'fs/promises';
+import { writeFileSync } from 'fs';
+import { readFile, writeFile } from 'fs/promises';
 import path from 'path';
 
 import { HTTPSnippet, Request } from '../httpsnippet';
@@ -21,17 +22,21 @@ export interface CustomFixture {
 export const runCustomFixtures = ({ targetId, clientId, tests }: CustomFixture) => {
   describe(`custom fixtures for ${targetId}:${clientId}`, () => {
     tests.forEach(({ it: title, expected: fixtureFile, options, input: request }) => {
+      const result = new HTTPSnippet(request).convert(targetId, clientId, options);
+      const filePath = path.join(
+        __dirname,
+        '..',
+        'targets',
+        targetId,
+        clientId,
+        'fixtures',
+        fixtureFile,
+      );
+      if (process.env.OVERWRITE_EVERYTHING) {
+        writeFileSync(filePath, String(result));
+      }
+
       it(title, async () => {
-        const result = new HTTPSnippet(request).convert(targetId, clientId, options);
-        const filePath = path.join(
-          __dirname,
-          '..',
-          'targets',
-          targetId,
-          clientId,
-          'fixtures',
-          fixtureFile,
-        );
         const buffer = await readFile(filePath);
         const fixture = String(buffer);
 
--- a/src/targets/shell/curl/client.test.ts
+++ b/src/targets/shell/curl/client.test.ts
@@ -1,3 +1,4 @@
+import applicationFormEncoded from '../../../fixtures/requests/application-form-encoded.json';
 import full from '../../../fixtures/requests/full.json';
 import nested from '../../../fixtures/requests/nested.json';
 import { runCustomFixtures } from '../../../fixtures/runCustomFixtures';
@@ -70,6 +71,21 @@ runCustomFixtures({
         indent: '@',
       },
       expected: 'custom-indentation.sh',
+    },
+    {
+      it: 'should url encode the params key',
+      input: {
+        ...applicationFormEncoded,
+        postData: {
+          mimeType: 'application/x-www-form-urlencoded',
+          params: [
+            { name: 'user name', value: 'John Doe' },
+            { name: '$filter', value: 'by id' },
+          ],
+        },
+      } as Request,
+      options: {},
+      expected: 'urlencode.sh',
     },
   ],
 });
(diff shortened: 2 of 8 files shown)


src/targets/shell/curl.js
```diff
@@ -58,10 +58,17 @@ module.exports = function (source, options) {
...
-          )
+          if (opts.binary) {
+            code.push(
+              '%s %s', '--data-binary',
+              helpers.quote(util.format('%s=%s', param.name, param.value))
+            )
+          } else {
+            code.push(
+              '%s %s', '--data-urlencode',
+              helpers.quote(util.format('=%s=%s', param.name, param.value))
```
erunion: I don't think this preceding `=` should be here.

```suggestion
              helpers.quote(util.format('%s=%s', param.name, param.value))
```

jgiovaresco: It was intended. To tell to cUrl to URL encode the data, we should use `--data-urlencode`. This will encode the value, but we should also encode the key. It can be done by prefixing the string to encode with `=`.

Extract from this [page](https://ec.haxx.se/http/http-post#url-encoding):

>If you want to URL encode the field name as well, like if you want to pass on a field name called "user name", you can ask curl to encode the entire string by prefixing it with an equals sign (that will not get sent):
```
curl --data-urlencode "=user name=John Doe (Junior)" http://example.com
```

jgiovaresco: It was intended to let cUrl URL encodes the data. I was reading this [page](https://ec.haxx.se/http/http-post#url-encoding):

> If you want to URL encode the field name as well, like if you want to pass on a field name called "user name", you can ask curl to encode the entire string by prefixing it with an equals sign (that will not get sent):

Reading it one more time has started to make me doubt 🤔 

The following

```
curl --request POST \
  --url http://localhost:3000/url-encoded \
  --header 'Content-Type: application/x-www-form-urlencoded' \
  --data-urlencode '=$filter=my&value = 1'
```

produces the encoded string: `%24filter%3Dmy%26value%20%3D%201` 
and this is badly interpreted in `{"$filter=my&value = 1":""}`

Removing the `=` will let cUrl to encode the value correctly. 
However, we should also encode the `key`. I've thought cUrl would be able to do it for us, but it seems it won't. Therefore I will encode the key "manually."

jgiovaresco:
I'm encoding the key using `encodeURIComponent` ([here](https://github.com/Kong/httpsnippet/pull/198/files#diff-dcda60e32cba1a3d324d44f7345c216d642944a140b0f0c58b261c38dc9472e2R63)).

I'm wondering if we should use it for the value or keep using `--data-urlencode` to let cUrl encode the value automatically. 🤔

bbbco:
I came across this issue as well, but I'd argue that we should only be url-encoding values that require quoting. For example, the curl snippet output should look something like this where the values that required quoting are the ones passed using `--data-urlencode`. 

```
curl --request POST \ 
  --url http://localhost/oauth/token \
  --header 'Accept: application/json' \
  --header 'Content-Type: application/x-www-form-urlencoded' \
  --data grant_type=password \
  --data client_id=XXXXXXXXXXXXXXX \
  --data-urlencode 'client_secret=sdhoisDDSHOIDHs/++SDOIhoDSHOIDHS' \ # made up
  --data username=foo \
  --data-urlencode 'password=P@$$w0rD'
```

This is working correctly for me on the CLI

dimitropoulos (approved):

dimitropoulos:
thanks @bbbco for the heads-up.  I was thinking the same exact thing when I was reviewing this.  I went ahead and fixed it like you've suggested in https://github.com/Kong/httpsnippet/pull/198/commits/f2c115545f15f3a96df44cfd2acfa4ecef7c876a.

src/targets/shell/curl.js
dimitropoulos: I also am not sure about this.  I think given that we don't know, the best case is to leave it "as it was" and see if that works for people.

I'll leave this unresolved so that if you, the reader (some time in the future), are experiencing this: you'll see what our reasoning was.  If this presents you a problem please open a GitHub issue!

PR #199, opened 2020-11-10
Add --globoff option to shell_curl

adamrdavid:
In some common api specifications there exist nested query parameters that take the form of an object.
Example: `page[limit]=10` https://jsonapi.org/

### This PR
1. Adds a fixture for nested query params and associated test cases
2. Adds the `--globoff` or `-g` option for the curl command to allow unescaped brackets in the url

### DIFF
--- a/src/targets/shell/curl.js
+++ b/src/targets/shell/curl.js
@@ -18,13 +18,22 @@ module.exports = function (source, options) {
   var opts = Object.assign({
     indent: '  ',
     short: false,
-    binary: false
+    binary: false,
+    globOff: false
   }, options)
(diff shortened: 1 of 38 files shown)


reynolek (changes requested):
With a recent change, we are converting over to using `const` and `let` rather than `var`. There is likely also a merge conflict with this exact issue as well. Would you mind sweeping through your new code and eliminating `var`, and then I can get this merged.

reynolek (approved):

reynolek:
Thanks for knocking that out! Merging now.

PR #202, opened 2021-01-30
fix: bug where FormData in the js fetch target wasn't applied

erunion:
* [x] Fixes a bug in the JS `fetch` target where `FormData` was being set to a `form` variable but never actually inserted. into the body of the fetch call.
* [x] Stops the JS fetch target from setting an empty object of `headers`.
* [x] Slightly cleans up the noise around the actual fetch call and indents the promise handling.

### DIFF
--- a/src/targets/javascript/fetch.js
+++ b/src/targets/javascript/fetch.js
@@ -67,7 +67,7 @@ module.exports = function (source, options) {
       }
   }
 
-  code.push('const options = %s;', stringifyObject(options, {indent: opts.indent, inlineCharacterLimit: 80}))
+  code.push('const options = %s;', stringifyObject(options, { indent: opts.indent, inlineCharacterLimit: 80 }))
     .blank()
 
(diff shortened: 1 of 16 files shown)


jbwyme:
Hey just wanted to check in on the status of this! I believe this is causing incorrect code to be generated in our ReadMe docs.

erunion:
@jbwyme You've already got this fix if you're a ReadMe customer because we run a fork (http://npm.im/@readme/httpsnippet) that have these changes. What's the bug you're seeing?

jbwyme:
Ah gotcha, perhaps this is unrelated. Here's the loom I shared with support: https://www.loom.com/share/06da93107f4b40f88172d680507ac99d

erunion:
@jbwyme Gotcha, that hit my desk yesterday.

reynolek:
Fixed merge conflict injected from the PR that removed node 8

reynolek (approved):

PR #203, opened 2021-01-30
feat: adding an option for disabling query string encoding

erunion:
We've discovered a bug where if your HAR definition has query strings that are already encoded (eg: `startTime: 2015-09-15T14%3A00%3A12-04%3A00` instead of `startTime: 2015-09-15T14:00:12-04:00`), the library will re-encode them again when it constructs the `urlObj` and `fullUrl` data that a handful of targets use. This leads to those parameters values getting corrupted.

This PR adds a new argument, and option, to disable this escaping that happens on `urlObj` and `fullUrl`:

```js
new HTTPSnippet(har, { escapeQueryStrings: true })
```

I've added a new `query-encoded` test case that covers this case for every available target.

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,6 +83,14 @@ var snippet = new HTTPSnippet({
 });
 ```
 
+#### options
+
+Type: `object`
+
+Available options:
+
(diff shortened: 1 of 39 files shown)


erunion:
Actually, nevermind!

PR #204, opened 2021-01-30
rfc: drop support for < Node 10

erunion:
The current earliest version of Node that HTTPSnippet targets is Node 4 which was EOL'd in 2018. Probably time to cut it loose.

* [x] Zero changes to library functionality or usage.
* [x] Changes the `engines` requirement to be >=10.
  * One thing that'll need to be addressed at some point is that `url.parse` was deprecated in Node 11, and `urlObj` is populated with that.
  * We might be able to eschew the `url` dependency all together since it's just being used for `url.parse` and `url.format`. The new [URL API](https://developer.mozilla.org/en-US/docs/Web/API/URL) should be able to be adequate.
* [x] Upgrades `mocha` and `standard` to their latest versions now that we're no longer blocked by the node@4 requirement.
* [x] Fixes a host of Standard issues that appeared with the upgrade.
* [x] Removes the `pinkie-promise` dependency as it's not used anywhere anymore.

### DIFF
--- a/.jshintrc
+++ b/.jshintrc
@@ -1,5 +1,0 @@
-{
-  "asi": true,
-  "browser": true,
-  "node": true
-}
--- a/.npmignore
+++ b/.npmignore
@@ -1,3 +1,2 @@
-.jshintrc
 .editorconfig
 test
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,6 @@ language: node_js
 node_js:
  - node
  - lts/*
- - 8
 
 before_install:
   - sudo apt-get update -qq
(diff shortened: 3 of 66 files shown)


reynolek (approved):
Thanks @erunion -- The change looks good. I am thinking of making this the `v2` release so that people arent forced to upgrade to node 10 (although they really should be *at least* node 10!).

I want to do 1 more minor release with some changes that have been merged, and then I will merge this and release `v2`

reynolek:
We will also need to remove a Node 8 build check in travis.

reynolek:
I resolved a conflict after merging another one of your PRs.

erunion:
@reynolek awesome ty

PR #205, opened 2021-02-08
feat: add parameter to enable decode url for place holder

lzyLuke:
Hi, forks.
I am experiencing some issues of using this library.
Say I have a url : http://mockbin.com/request/{id}, the {id} is a place holder. When generating snippets, I do not want the place holder to be uri encoded. 
If I use current code: there would be a  "http://mockbin.com/request/%7Bid%7D"
But I actually want "http://mockbin.com/request/{id}" here.

So I want to add a switch on httpSnippet to decide whether we should use uriencode here. (Default is yes, so this will not be a breaking change)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 *.log
 node_modules
 coverage*
+.DS_Store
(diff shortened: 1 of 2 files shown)


reynolek:
Hey @lzyLuke! This looks like a useful change. Sorry it took me so long to get to it!

reynolek (changes requested):
The change itself looks fine to me, but would you mind adding some test coverage so that Future Us doesn't unintentionally break this functionality?

dimitropoulos:
see https://github.com/Kong/httpsnippet/pull/290/

PR #210, opened 2021-03-30
update javascript fetch request to match node-fetch

jonasholbech:
Vanilla JS did not return the json from the request, the node.js version does, now they should behave the same


erunion (changes requested):
Can you update the tests for the JS target? They need this line added to them.

jonasholbech:
Let me see if I can find those :-)

erunion:
Here you go: https://github.com/Kong/httpsnippet/tree/master/test/fixtures/output/javascript/fetch

jonasholbech:
thanks @erunion

reynolek (approved):

reynolek:
Thanks for the contribution @jonasholbech!

PR #213, opened 2021-04-20
fix: node-fetch doesn't support the `qs` option

### LINKED ISSUE #201: Node.js / Fetch - `qs` isn't supported
Code Generation for Node.js / Fetch using `node-fetch` generates code like this:

```

let url = '<REDACTED>';

let options = {
  method: 'GET',
  qs: {reference: '<REDACTED>'},
  headers: {
    'Content-Type': 'application/json'
  }
};

fetch(url, options)
  .then(res => res.json())
```

However the `qs` option is not supported by `node-fetch`: https://github.com/node-fetch/node-fetch/issues/887

erunion:
This fixes the Node `fetch` target to no longer supply query params through the `qs` option (which it does not support).

Fixes https://github.com/Kong/httpsnippet/issues/201

### DIFF
--- a/src/targets/node/fetch.js
+++ b/src/targets/node/fetch.js
@@ -22,13 +22,9 @@ module.exports = function (source, options) {
   var code = new CodeBuilder(opts.indent)
 
   code.push('const fetch = require(\'node-fetch\');')
-  var url = source.url
+  var url = source.fullUrl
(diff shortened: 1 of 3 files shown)


reynolek (approved):

reynolek:
Thanks as always @erunion !

PR #216, opened 2021-05-11
Moved the post data for the php curl target into its own variable and utilized http_build_query() so we can work with a nicely formatted array instead of  that long string

detain:
Moved the post data for the php curl target into its own variable and utilized http_build_query() so we can work with a nicely formatted array instead of  that long string

Before Change Snippet:
```
<?php

$curl = curl_init();

curl_setopt_array($curl, [
  CURLOPT_URL => "https://api.mailbaby.net/mail/send",
  CURLOPT_RETURNTRANSFER => true,
  CURLOPT_ENCODING => "",
  CURLOPT_MAXREDIRS => 10,
  CURLOPT_TIMEOUT => 30,
  CURLOPT_HTTP_VERSION => CURL_HTTP_VERSION_1_1,
  CURLOPT_CUSTOMREQUEST => "POST",
  CURLOPT_POSTFIELDS => "to=hi&from=there&subject=something&body=here",
  CURLOPT_HTTPHEADER => [
    "Content-Type: application/x-www-form-urlencoded",
  ],
]);
```

After Change Snippet:
```
<?php

$postData = http_build_query([
  'to' => 'hi',
  'from' => 'there',
  'subject' => 'something',
  'body' => 'here'
]);

$curl = curl_init();

curl_setopt_array($curl, [
  CURLOPT_URL => "https://api.mailbaby.net/mail/send",
  CURLOPT_RETURNTRANSFER => true,
  CURLOPT_ENCODING => "",
  CURLOPT_MAXREDIRS => 10,
  CURLOPT_TIMEOUT => 30,
  CURLOPT_HTTP_VERSION => CURL_HTTP_VERSION_1_1,
  CURLOPT_CUSTOMREQUEST => "POST",
  CURLOPT_POSTFIELDS => $postData,
  CURLOPT_HTTPHEADER => [
    "Content-Type: application/x-www-form-urlencoded",
  ],
]);
```


detain:
Updated the code so it passes all the linter and tests now, as well as fixing a few problems I discovered through the testing

test/fixtures/output/php/curl/multipart-file.php
```diff
@@ -1,5 +1,7 @@
 <?php
 
+$postData = "-----011000010111000001101001\r\nContent-Disposition: form-data; name=\"foo\"; filename=\"hello.txt\"\r\nContent-Type: text/plain\r\n\r\n\r\n-----011000010111000001101001--\r\n";
```
erunion: Thoughts on introducing some splitting on `\n` to transform this line into something like the following?

```php
$postData = "-----011000010111000001101001\r\n";
$postData += "Content-Disposition: form-data; name=\"foo\"; filename=\"hello.txt\"\r\n";
$postData += "Content-Type: text/plain\r\n\r\n\r\n-----011000010111000001101001--\r\n";
```

erunion (approved):
Have a small suggestion, but it might be out of scope for your work here. Looks great though!

test/fixtures/output/php/curl/multipart-file.php
detain: I think some changes along those lines would be good; but should be saved for a different pull (that and I don't have a great solution in my head at this moment).

PR #219, opened 2021-05-13
feat: cleaner json usage within curl snippets

erunion:
This slightly modifies the cURL snippet target to clean up how it handles JSON payloads into rendering them on multiple lines.

This is what they currently look like:

```shell
curl --request POST \
  --url http://mockbin.com/har \
  --header 'content-type: application/json' \
  --data '{"number":1,"string":"f\"oo","arr":[1,2,3],"nested":{"a":"b"},"arr_mix":[1,"a",{"arr_mix_nested":{}}],"boolean":false}'
```

With this change:

```shell
curl --request POST \
  --url http://mockbin.com/har \
  --header 'content-type: application/json' \
  --data '
{
  "number": 1,
  "string": "f\"oo",
  "arr": [
    1,
    2,
    3
  ],
  "nested": {
    "a": "b"
  },
  "arr_mix": [
    1,
    "a",
    {
      "arr_mix_nested": {}
    }
  ],
  "boolean": false
}
'
```

## 🐳  Things to note
* I don't love how arrays and objects get blown out if they're small enough to be able to fit on a single line (like `arr` is above), but the [stringify-object](https://www.npmjs.com/package/stringify-object) library that's used throughout this library for doing prettification of objects removes quotes from object keys, which in this case would result in invalid JSON from being sent.
* If the JSON payload has a `'` quote within it, we can't wrap that in a `--data` string without busting STDIN so instead we wrap the JSON within a HEREDOC:
 
```shell
curl --request POST \
  --url http://mockbin.com/har \
  --header 'content-type: application/json' \
  --data @- <<EOF
{
  "number": 1,
  "string": "f'oo"
}
EOF
```

* To prevent miniscule JSON payloads from getting blown out snippets, if the stringified JSON is less than or equal to 20 characters we'll render it on one line:

```shell
curl --request POST \
  --url http://mockbin.com/har \
  --header 'content-type: application/json' \
  --data '{"foo":null}'
```


reynolek:
@erunion I was going to merge this for the next release, but with the conflicts I didnt feel confident fixing them up for you (happened due to the change from #227)

Mind fixing it up and ill merge it later.

erunion:
@reynolek I just fixed the merge conflict so this should be good to go after Travis gives it the 👍

src/helpers/headers.js
```diff
@@ -35,5 +35,23 @@ module.exports = {
...
+   * @returns {boolean}
+   */
+  isMimeTypeJson: (mimeType) => {
+    return [
+      'application/json',
+      'application/x-json',
+      'text/json',
+      'text/x-json',
+      '+json'
+    ].some(function (type) {
```
ghost, line 53: NOTE: you could condense lines 53-55 to:
```
      ].some(mimeType.includes.bind(mimeType))
```

PR #222, opened 2021-05-27
pass extra options to httpsnippet (pass through to the target)

mahenzon:
Hi! I could not find an option to provide extra options to the target. For example if I don't need Host or ContentLength at the http target. Added an option to provide extra options to the target, updated readme

Example:

```shell
httpsnippet example.json --target http --output ./snippets -x '{"autoHost": false, "autoContentLength": false}'
```

### DIFF
--- a/bin/httpsnippet
+++ b/bin/httpsnippet
@@ -17,10 +17,21 @@ cmd
   .option('-t, --target <target>', 'target output')
   .option('-c, --client [client]', 'target client library')
   .option('-o, --output <directory>', 'write output to directory')
+  .option('-x, --extra [{"optionKey": "optionValue"}]', 'provide extra options for the target/client')
   .parse(process.argv)
 
 if (!cmd.args.length || !cmd.target) {
   cmd.help()
+}
+
+let extraOptions
+if (cmd.extra) {
+  try {
+    extraOptions = JSON.parse(cmd.extra)
+  } catch (e) {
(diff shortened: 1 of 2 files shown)


bin/httpsnippet
```diff
@@ -17,12 +17,23 @@ cmd
   .option('-t, --target <target>', 'target output')
   .option('-c, --client [client]', 'target client library')
   .option('-o, --output <directory>', 'write output to directory')
+  .option('-x, --extra [{"optionKey": "optionValue"}]', 'provide extra options for the target/client')
```
erunion, line 20: Thoughts on having this so that any argument not strictly defined on the CLI would be interpreted as an option for the targets and be automatically bundled up into `extraOptions`? Like:

```
httpsnippet example.json --target http --output ./snippets --autoHost false --autoContentLength false
```

mahenzon: Yeah, I had a thought about it too. But it'll take a lot more effort to handle it and parse correctly.  Is there only true/false, or other options are available like strings or maybe even arrays? We'll need to parse each option and convert to to the correct type. Is it worth it?

mahenzon: @erunion what does your '+1' mean? is it "ok, really no need to spend time on parsing, passing JSON config is ok", or "yeah you better work on  passing options in `--autoHost false --autoContentLength false` format"?

erunion: It's "that makes sense, leaving it as it is and passing JSON is sounds good to me!"

erunion (approved):

reynolek (approved):

PR #224, opened 2021-07-09
feat: make Python snippets simpler, clearer & more consistent

pimterry:
This PR does three things (split into 3 commits):

* Previously the Python snippets used a mixture of single & double quotes, with a mixture of dictionary literal formats: e.g. `{'a':"b"}` vs `{ "a": "b" }`. It's now consistent, so there's spaces and nice formatting for all dictionary literals, and double quotes everywhere (that seemed to be the smaller change, Python community doesn't seem to have a strong preference on single/double, other than 'be consistent').
* Python Requests has built-in methods for most standard HTTP methods. The code now simplifies code by using those where possible, changing `requests.request("POST", ...)` for `requests.post(...)`.
* Python Requests can accept dictionary literals for form-encoded bodies, so we now do that when `source.postData.paramsObj` is available, to make that body content more readable.

### DIFF
--- a/src/targets/python/helpers.js
+++ b/src/targets/python/helpers.js
@@ -25,6 +25,8 @@ function concatValues (concatType, values, pretty, indentation, indentLevel) {
 
   if (pretty) {
     return openingBrace + '\n' + currentIndent + values.join(join) + '\n' + closingBraceIndent + closingBrace
+  } else if (concatType === 'object' && values.length > 0) {
+    return openingBrace + ' ' + values.join(join) + ' ' + closingBrace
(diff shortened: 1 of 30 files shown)


erunion (approved):

reynolek (approved):
LGTM -- Thanks for the contribution!

pimterry:
Is there anything else that needs to happen here before this can be merged?

pimterry:
@reynolek this is double approved and ready to go - is it possible to get it merged?

dimitropoulos (approved):

PR #227, opened 2021-09-03
fix(curl) remove boundary form content-type header

### LINKED ISSUE #226: curl multipart/form-data snippet includes unwanted boundary
The snippet generated for a multipart form data request for curl includes a boundary in the headers. curl appends it own boundary header, resulting in two boundaries and the request fails.

Example snippet generated (with the addition of the verbose flag): 
```
curl --request POST \
  --url https://mockbin.org/echo \
  --header 'Content-Type: multipart/form-data; boundary=---011000010111000001101001' \
  --header 'accept: application/json' \
  --form 'filename=@nyancat32x32.png;type=image/png' \
  -v
 ``` 
 
 We can see that the request was sent with double boundaries, one from httpsnippet, and one added by curl:
 ```
  Content-Type: multipart/form-data; boundary=---011000010111000001101001; boundary=------------------------c18ea9ee9be9cb74
  ```
  As far as I can tell, httpsnippet always adds the `boundary=---011000010111000001101001'` to every multipart/form-data request. Any ideas of a way around this?

### PR DESCRIPTION
**Pull request by nateslo**
Resolves https://github.com/Kong/httpsnippet/issues/226

This PR removes the boundary from the content-type header for curl multipart request. Curl appends its own boundary when using `--form`, resulting in a double boundary.

### DIFF
--- a/src/targets/shell/curl.js
+++ b/src/targets/shell/curl.js
@@ -12,6 +12,7 @@
 
 const util = require('util')
 const helpers = require('../../helpers/shell')
+const headerHelpers = require('../../helpers/headers')
 const CodeBuilder = require('../../helpers/code-builder')
 
 module.exports = function (source, options) {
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: src/targets/shell/curl.js [hunk 1]
```diff
@@ -39,6 +40,22 @@ module.exports = function (source, options) {
     code.push(opts.short ? '-0' : '--http1.0')
   }
 
+  // if multipart form data, we want to remove the boundary
```
reynolek, line 43: Per #226, we discussed just scoping this to fixing cURL due to how cURL handles `--form` vs raw data.

Did you attempt any kind of audit of other targets to see if they are impacted by similar issues?

src/targets/shell/curl.js
```diff
@@ -39,6 +40,22 @@ module.exports = function (source, options) {
...
+  if (source.postData.mimeType === 'multipart/form-data') {
+    const contentTypeHeaderName = headerHelpers.getHeaderName(source.headersObj, 'content-type')
+    const contentTypeHeader = source.headersObj[contentTypeHeaderName]
+
+    if (contentTypeHeaderName && contentTypeHeader) {
+      // remove the leading semi colon and boundary
+      // up to the next semi colon or the end of string
+      const noBoundary = contentTypeHeader.replace(/; boundary.+?(?=(;|$))/, '')
```
reynolek, line 51: I double checked these regex and functions as expected.

Screen Shot 2021-09-03 at 11 16 37 AM

reynolek (approved):

src/targets/shell/curl.js
nateslo: I have not investigate any other targets yet, this fix only pertains to curl multipart

erunion (approved):
big oof. thanks for fixing this.

if other targets need this boundary removed it might be worth refactoring this then into a part of the target config.

nateslo:
> big oof. thanks for fixing this.
> 
> if other targets need this boundary removed it might be worth refactoring this then into a part of the target config.

No worries! Thanks for the review

PR #230, opened 2021-09-15
fix: multipart/form-data header issues with node/js fetch targets

erunion:
Since #227 has us a bit spooked we've started testing other targets that may also have issues with `multipart/form-data` requests and found that both of the `fetch` targets for Node and JS are not currently transferring data properly because the custom header that is being added is clashing with the auto-generated header from the `FormData` API and `form-data` module.

You can see this problem in action with the `multipart-form-data` snippet against https://httpbin.org/anything:

```js
const FormData = require('form-data');
const formData = new FormData();

formData.append('foo', 'bar');

let url = 'https://httpbin.org/anything';

const options = {
  method: 'POST',
  headers: {'Content-Type': 'multipart/form-data; boundary=---011000010111000001101001'}
};

options.body = formData;

fetch(url, options)
  .then(res => res.json())
```

```js
{
  args: {},
  data: '',
  files: {},
  form: {},
  headers: {
    Accept: '*/*',
    'Accept-Encoding': 'gzip,deflate',
    'Content-Length': '161',
    'Content-Type': 'multipart/form-data; boundary=---011000010111000001101001',
    Host: 'httpbin.org',
    'User-Agent': 'node-fetch/1.0 (+https://github.com/bitinn/node-fetch)',
    'X-Amzn-Trace-Id': 'Root=1-61428408-5a6abe107e16c34466e8e6ba'
  },
  json: null,
  method: 'POST',
  origin: 'REDACTED',
  url: 'https://httpbin.org/anything'
}
```

After removing the `multipart/form-data` header:

```js
{
  args: {},
  data: '',
  files: {},
  form: { foo: 'bar' },
  headers: {
    Accept: '*/*',
    'Accept-Encoding': 'gzip,deflate',
    'Content-Length': '161',
    'Content-Type': 'multipart/form-data;boundary=--------------------------635295776437093815404580',
    Host: 'httpbin.org',
    'User-Agent': 'node-fetch/1.0 (+https://github.com/bitinn/node-fetch)',
    'X-Amzn-Trace-Id': 'Root=1-6142841e-4c789faa2a1d423452674f13'
  },
  json: null,
  method: 'POST',
  origin: '98.42.248.146',
  url: 'https://httpbin.org/anything'
}
```

### DIFF
--- a/src/targets/javascript/fetch.js
+++ b/src/targets/javascript/fetch.js
@@ -10,7 +10,7 @@
 
 'use strict'
 
-const headerHelpers = require('../../helpers/headers');
+const headerHelpers = require('../../helpers/headers')
 const CodeBuilder = require('../../helpers/code-builder')
 const stringifyObject = require('stringify-object')
 
@@ -32,10 +32,10 @@ module.exports = function (source, options) {
   // The `form-data` library automatically adds a `Content-Type` header for `multipart/form-data` content and if we
   // add our own here, data won't be correctly transferred.
   if (source.postData.mimeType === 'multipart/form-data') {
-    const contentTypeHeader = headerHelpers.getHeaderName(source.allHeaders, 'content-type');
+    const contentTypeHeader = headerHelpers.getHeaderName(source.allHeaders, 'content-type')
     if (contentTypeHeader) {
       // eslint-disable-next-line no-param-reassign
-      delete source.allHeaders[contentTypeHeader];
+      delete source.allHeaders[contentTypeHeader]
     }
   }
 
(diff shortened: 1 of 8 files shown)


src/targets/javascript/fetch.js
```diff
@@ -21,30 +23,39 @@ module.exports = function (source, options) {
     options
   )
 
-  const stringifyObject = require('stringify-object')
   const code = new CodeBuilder(opts.indent)
 
-  options = {
```
erunion, line 27: I renamed this variable to `reqOpts` so it's not overwriting the `options` argument of the target.

reynolek (approved):
Thanks @erunion -- I figured there might be a few other places multipart was broken.

kubasaw (approved):

PR #231, opened 2021-09-15
fix: bug in the node request target where cookies weren't sent at all

erunion:
This fixes a typo in some string replace code within the Node `request` target that was causing cookies to not be sent.


erunion:
Closing this out because the issues been fixed in `master` now.

PR #239, opened 2021-12-08
Selection of test setup improvements

pimterry:
(Last PR today I promise! Sorry for all the noise)

This PR fixes a few issues I ran into in putting together the other two:

* It provides a clear error when a test dependency is missing or fails. In my case, PHP wasn't installed, but previously all I saw was every PHP test failing and:
  ```
  Uncaught AssertionError: expected SyntaxError { message: 'Unexpected end of JSON input' } to be null
  ```
  Now, if PHP isn't installed, you get:
  ```
  Uncaught AssertionError: Process 'php' exited unexpectedly with code 127
  ```
* It removes the Travis config that's doing nothing here (because Travis is no longer doing free open source builds) and replaces it with an equivalent GitHub Actions config that works
* It updates the coverage scripts so that they work too. The only thing required to make this upload to Codeclimate correctly is to get the test reporter id (Codeclimate project -> Repo Settings -> Test Coverage -> Test reporter id) and set it as a secret called `CODECLIMATE_REPO_TOKEN` in this repo.

You can see a passing test build of this in the actions on my repo here: https://github.com/httptoolkit/httpsnippet/actions/runs/1555694270


dimitropoulos:
hey @pimterry!  thankfully this was implemented in https://github.com/Kong/httpsnippet/pull/248.  I hadn't seen your solution so the way it was done there is slightly different, but suffice to say: yes! I totally agree that it left a lot to be desired before.

Take a look at the new implementation here: https://github.com/Kong/httpsnippet/blob/d4bebe4ae27712021e94f8458064fe7c8599d493/src/targets/targets.test.ts#L57

PR #246, opened 2022-02-17
Add %v and %q options to format in CodeBuilder.buildLine and some handy indentation methods

uWynell:
1. Added `%v` and `%q` options to format in `CodeBuilder.buildLine`. Why? When I implemented Rust's Reqwest target (I'll make the corresponding Pull Request after this one), I noticed the problem of escaping characters. For example, my chrome browser generates `sec-ch-ua` header, that contains quotes, and some of the targets do not escape them, which results in broken code. And also I'm using it in the mentioned Reqwest target.
2. Added `indent`, `unindent` and `reindent` methods to `CodeBuilder`, as it allows to control indentation more conveniently.

### How do `%v` and `%q` options work?
Right from the comments I made to the new `format` function:

> Wraps the util.format function and adds the %q and %v format options, where %q - escape tricky characters, like newline or quotes and %v - JSON-stringify-if-necessary

Examples:
```js
format('foo("%q")', { bar: 'baz' })
// output: foo("{\"bar\":\"baz\"}")

format('foo("%q")', `
`)
// output: foo("\n")

format('foo(%v)', { bar: 'baz' })
// output: foo({"bar":"baz"})

format('foo(%s)', { bar: 'baz' })
// output: foo({ bar: 'baz' })
```


uWynell:
Sorry, I didn't notice that I accidentally changed package-lock when I didn't have to

dimitropoulos:
hey @uWynell!  sorry that I didn't see this sooner, but in https://github.com/Kong/httpsnippet/pull/248 we did a bit of a total overhaul and vastly simplified this stuff (in a slightly different way).  I'm going to close this PR for now in part also because we removed the `utils` package usages from the entire package and won't be adding it back or expanding upon it.

Really appreciate your PR, truly, and hope to see more in the future!  If I mixed something up please let me know!

PR #247, opened 2022-02-17
add header namesspace to prevent header errors

davismartin:
## Description 

Currently, if I try to just run the snippets produced by this charp -> httpclient.js target then I will get errors similar to 
```
error CS0246: The type or namespace name 'ContentDispositionHeaderValue' could not be found
```

However, if I add the namespace `using System.Net.Http.Headers;` to the top then the snippet works as expected

This PR just adds that namespace to the top so that the snippet can run if just copied.

changelog(Fixes): Fixed an issue where C# snippets were missing a namespace in the generated code

### DIFF
--- a/src/targets/csharp/httpclient/client.ts
+++ b/src/targets/csharp/httpclient/client.ts
@@ -48,6 +48,7 @@ export const httpclient: Client = {
 
     const { push, join } = new CodeBuilder({ indent: opts.indent });
 
+    push('using System.Net.Http.Headers;');
     let clienthandler = '';
     const cookies = Boolean(allHeaders.cookie);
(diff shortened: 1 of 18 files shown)


erunion (changes requested):
Can you update the test fixtures with this new line?

erunion (approved):

dimitropoulos (approved):
(I rebased it into TypeScript-land) looks good to me!  I wondered if it'd be more nature to have a new line after the `using` statement.  but all good.  we can easily add later to tweak if so.  figured I'd default to someone that's clearly familar with csharp these days (@davismartin)

PR #248, opened 2022-04-08
Total Overhaul (but with all the same fixtures!)

### LINKED ISSUE #186: Modernize codebase: TypeScript migration 🙈
Codebase might seem quite outdated: `vars` instead of `const`/`let`, old way to define a class

It could be nice to modernize it. Currently, the `package.json` is telling us that we could use the library with at least Node 4. Therefore we can't just update the source code because we should keep it compatible with old Node versions.

I would suggest configuring Babel or TypeScript to be able to use the latest ES standard and stay compatible with the old node versions.

dimitropoulos:
## Background

Hi!

I'm Dimitri, and I am one of the developers here at Kong working (primarily) on Insomnia.  I realize it may not have been clear in the past, but at Kong we consider my team to be the "owners" of httpsnippet.

One of our core values here at Kong is "Real", so I'm gonna just keep it real and say that I realize that this package has been largely abandoned (for the most part) over the last few years.  Especially the last year, where there has been plenty of pull requests and issues and almost none were even answered by any maintainers.

The thing is... we feel this pain on Insomnia too.  For example, we were looking at some open source issues on our [weekly Insomnia livecoding stream](https://youtu.be/mueVITe2quw?t=1129) and you can just about see the moment when the fun drains out of the room when we realize that to modify a behavior in Insomnia, we're going to have to touch httpsnippet.  Yikes.

Thankfully, this wasn't the first time this library had come up recently and a few different people at Kong have been running into problems with this library.  And to be fair, it's not only "problems" that we hit, we also have aspirations for how we could leverage this tool better in the future. So, beginning about a week ago, I started taking some time to look at this repo and understand how it worked after work hours.  I quickly concluded that there are two very VERY positive things we have going for us:
1. there are TONS and TONS (I mean like SIX HUNDRED) of fixture tests for all the various clients
1. the problem-domain of this package is very finite.  it's a "solveable" problem (in terms of the functionality the package provides).

So, then, I set about the job of retooling it from the ground up.  It's a completely modernized stack.  Many of the same nuts and bolts, but now using things like template literals (instead of those ancient and error-prone utils.format magic strings), destructuring, optional chaining, etc. and all tied together with TypeScript in full strict mode!  That's not to mention it runs tests in Jest, uses Yargs, Yarn, lots more (and modern) linting, prettier, and we were able to throw out a good few dependencies.

## What to do with this PR

This is going to be a major release, for sure, but it's also going to be a pretty hefty breath of life into this project.

Immediately after this PR there are a few things I'd like to do, yet, but I'm making this PR now because I think what's here can merge.  I'd like to do the CI with GitHub Actions, for example, and fix the dockerfiles, use jest to report line coverage, etc.

I'd like to do a few betas, maybe, after this releases so we can dogfood this in Insomnia and make sure it's ready for the world.

Here are some examples:

Screenshot_20220408_173508

Screenshot_20220408_165313

## What changed

Check out TODO.txt on the root in this branch to see a list I've been keeping as I go of what _had_ to change.  But the quick answer is: _not much_ of any real significance.

## How do I review this

The key key key point I want you to know before you do any reviewing is that I didn't touch any of the main input-output fixture tests.  This is **really important** because it means that, at a minimum, the primary behavior (_to the character!_) of the package has not changed.

I don't want to make light of how much was moved around.  If you were familiar with this package before, you'll immediately see that things are organized pretty differently.  For example, instead of having 3 places in the repo where the targets/clients structure was duplicated (source code, unit tests, and fixture tests) it's all co-located in the same place.

I kept the commits there for (mainly) my own reference _ONLY_.

**PLEASE** DO NOT feel the need to look commit by commit.  Everything changed (except the raw fixtures!) so it's most definitely best to just look at it fresh.

## What's the plan going forward

I'm super open to feedback!

From the Insomnia team I'm gonna invite @filfreire @DMarby @johnwchadwick @davidma415 @gatzjames @jackkav and @wdawson to take a look and let me know if they have any thoughts.  No pressure.

Beyond that, there are a few others that I would love to hear from including @thibaultcha, @darrenjennings, @reynolek, @jeffyongtaotang, @gschier, @nateslo, @develohpanda but I don't have the expectation that any of you will respond, no worries!

Then, I'd like to recognize @erunion.  You clearly care about this project and I'd love to talk with you about what we can do to bring you a little more into the fold (i.e. the ability to review and merge PRs.  Could you please reach out to me over on the [insomnia slack](chat.insomnia.rest): `@Dimitri Mitropoulos`.

_____

I'm sure there's lots more to say, but I have a pizza in the oven and I'm hungry, haha.  It's been a long (and rewarding!) day so I'm gonna stop there for now.  Please feel free to ask any questions (anyone!).  Hopefully a bright future to come!

closes: #186


TODO.txt
```diff
@@ -0,0 +1,21 @@
+TODO:
+- convert to GitHub actions
+- fix Dockerfile and docker-compose
+- look into the prior usage of the 'debug' package
+- look into CLI and CLI tests
+  - also, move to yargs from commander
+- completely redo the README
+
+later:
+- use prettier for js languages (and more?)
```
erunion: For what it's worth we've looked into this at [ReadMe](http://readme.com/) and the cost of including Prettier in the bundles for this to work is going to be so prohibitiely expensive that it hasn't been worth it.

TODO.txt
```diff
@@ -0,0 +1,21 @@
...
+- add option to curl for minifying json
+
+changelog:
+
+## Breaking
+- drops support for node versions less than 16.13
```
erunion: Curious but what's the reasoning for not supporting Node 14? It's still in LTS until next April.

src/targets/powershell/restmethod/fixtures/echo
```diff
@@ -0,0 +1,16 @@
+$headers=@{}
```
erunion: I don't think this fixture is used anywhere.

src/fixtures/customTarget.ts
```diff
@@ -0,0 +1,14 @@
+import { request } from '../targets/node/request/client';
+import { Target } from '../targets/targets';
+
+export const customTarget = {
+  info: {
+    key: 'js-variant',
```
erunion, line 6: If you change this to `javascript` you shouldn't need to cast this object as `unknown` before `Target`.

README.md
```diff
@@ -70,135 +67,134 @@ provide extra options:
...
-Type: `object`
+_Required_ Type: `object`
 
 Representation of a [conversion target](https://github.com/Kong/httpsnippet/wiki/Creating-Targets). Can use this to use targets that are not officially supported.
 
-```js
-const customLanguageTarget = require('httpsnippet-for-my-lang');
+```ts
+import { customLanguageTarget } from 'httpsnippet-for-my-lang';
 HTTPSnippet.addTarget(customLanguageTarget);
```
erunion, line 171: These APIs have changed slightly with the new `addTarget` and `addTargetClient` exports on the main entrypoint.

src/fixtures/runCustomFixtures.ts
```diff
@@ -0,0 +1,40 @@
...
+
+import { HTTPSnippet, Request } from '../httpsnippet';
+import { ClientId, TargetId } from '../targets/targets';
+
+/* eslint-disable jest/no-export,jest/valid-title -- we want to do it just for this one case */
+export interface CustomFixture {
+  targetId: TargetId;
+  clientId: ClientId;
+  tests: {
+    title: string;
```
erunion: Thoughts on renaming this to `it` so it's more Jest-y?

src/helpers/code-builder.ts
```diff
@@ -0,0 +1,67 @@
+const DEFAULT_INDENTATION_CHARACTER = '';
```
erunion, line 1: The CodeBuilder no longer requiring the `util` module is awesome!

src/helpers/form-data.ts
```diff
@@ -0,0 +1,75 @@
...
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ *
+ * Extracted from https://github.com/node-fetch/node-fetch/blob/64c5c296a0250b852010746c76144cb9e14698d9/src/utils/form-data.js
+ */
+
+import FormData from 'form-data';
```
erunion: This is more of a larger topic of discussion since it'd involve updating a few other files that do this, but since import of `form-data` is only being used for the `FormData` type thoughts on the `import type` syntax?

```ts
import type FormData from 'form-data';
```

src/targets/php/http2/client.ts
```diff
@@ -0,0 +1,152 @@
...
+          [anything: string]: string | undefined;
+        }[] = [];
+        const fields: Record<string, any> = {};
+        postData.params?.forEach(({ name, fileName, value, contentType }) => {
+          if (fileName) {
+            files.push({
+              name,
+              type: contentType,
+              file: fileName,
+              // @ts-expect-error appears to be a genuine error
```
erunion: What's the error here?

src/targets/python/requests/client.test.ts
```diff
@@ -0,0 +1,17 @@
...
+  targetId: 'python',
+  clientId: 'requests',
+  tests: [
+    {
+      fixtureFile: 'query-params.py',
+      options: {
+        showBoilerplate: false,
+      },
+      request: { method: 'GET', url: 'http://mockbin.com/har?param=value' } as Request,
+      title: "should support query parameters provided in HAR's url",
```
erunion: Want to move this to the top of the test object so it's consistent with how the others are set up?

src/targets/shell/curl/client.ts
```diff
@@ -0,0 +1,134 @@
...
+    const opts = {
+      indent: '  ',
+      short: false,
+      binary: false,
+      globOff: false,
+      ...options,
+    };
+    const { push, join } = new CodeBuilder({
+      indent: opts.indent,
+      // @ts-expect-error SEEMS LEGIT
```
erunion: What's the error here?

src/targets/shell/curl/client.ts
```diff
@@ -0,0 +1,134 @@
...
+      join: opts.indent !== false ? ` \\\n${opts.indent}` : ' ',
```
erunion, line 40: The typing error on this one is definitely strange. Could maybe do an `opts.indent.length` check instead to resolve it?

src/targets/shell/curl/client.ts
```diff
@@ -0,0 +1,134 @@
...
+    // if multipart form data, we want to remove the boundary
+    if (postData.mimeType === 'multipart/form-data') {
+      const contentTypeHeaderName = getHeaderName(headersObj, 'content-type');
+      if (contentTypeHeaderName) {
+        const contentTypeHeader = headersObj[contentTypeHeaderName];
+        if (contentTypeHeaderName && contentTypeHeader) {
+          // remove the leading semi colon and boundary
+          // up to the next semi colon or the end of string
+          // @ts-expect-error SEEMS LEGIT
+          const noBoundary = contentTypeHeader.replace(/; boundary.+?(?=(;|$))/, '');
```
erunion, line 67: Also weird that `contentTypeHeader` might be a `string[]` sometimes.

src/httpsnippet.ts
```diff
@@ -0,0 +1,327 @@
...
+          // Since the polyfill does not support the full API of the native FormData object, when this library is running in a browser environment it'll fail on two things:
+          //
+          //  1. The API for `form.append()` has three arguments and the third should only be present when the second is a
+          //    Blob or USVString.
+          //  1. `FormData.pipe()` isn't a function.
+          //
+          // Since the native FormData object is iterable, we easily detect what version of `form-data` we're working with here to allow `multipart/form-data` requests to be compiled under both browser and Node environments.
+          //
+          // This hack is pretty awful but it's the only way we can use this library in the browser as if we code this against just the native FormData object, we can't polyfill that back into Node because Blob and File objects, which something like `formdata-polyfill` requires, don't exist there.
+          // @ts-expect-error TODO
```
erunion, line 189: Yeah... I'm sorry about this. I made an attempt to type this same thing in some stuff I'm working on in an unrelated project and pretty quickly gave up. I wish working with these FormData polyfills wasn't such a disaster or required ESM codebase migrations.

erunion (approved):
This is a great refactor! I've long wanted to rework this over into TS because so much of the individual options handling between each target has always felt really messy and am glad that you fell on that sword for all of us.

No longer requiring the `util` Node module is a huge benefit to web bundle compilation sizes. I **love** the new directory structure and containing target fixtures/expectations alongside the target. This should help to really make it easier add new targets/clients and reduce the PR noise of walking someone through contributing fixtures.

I appreciate your comments about my contributions to the project over the past couple years and will definitely reach out to you.

Awesome work! <sub>(and sorry for all the review comments 😅 )</sub>

TODO.txt
dimitropoulos: that sounds right.  welp.  maybe if nothing else we can have a devDependency to check that we're adhering to prettier formatting.

TODO.txt
dimitropoulos: yep. totally.  we'll go with 14.19. good catch.

src/targets/powershell/restmethod/fixtures/echo
dimitropoulos: great catch

src/fixtures/runCustomFixtures.ts
dimitropoulos: great feedback, updated to this in 59de6cd889bb60838da3ba198f13829d15ddee5e

TODO.txt
dimitropoulos: fixed in df36e95f24af3a329a946090e3614a8b97e1d6a5

src/fixtures/customTarget.ts
dimitropoulos: see https://youtu.be/5Di4bmIGTrg?t=3197 for an explanation for the approach taken here and lemme know (unresolve) if you have any better idea to solve the same problem without having to specify the key every time.

README.md
dimitropoulos: see the response here https://youtu.be/5Di4bmIGTrg?t=3447 but I _think_ I hold the position that this hasn't, in fact, changed for _THESE_ exact functions.

If you had done
```ts
const { addTarget } = require('httpsnippet');
```

before, that would still be valid today.  The real change was to how the default worked, which influenced how it was used before.  please unresolve if I missed something! :)

src/helpers/code-builder.ts
dimitropoulos: Thanks!  Yeah, I didn't actually set out with that goal, but we ended up achieving it by the simple fact of just using regular template strings.

src/helpers/form-data.ts
dimitropoulos: sure d0761ce99e5c88a4a0fc41b43f4a9ff1484d9d6c

src/targets/php/http2/client.ts
dimitropoulos: good catch: I think when I ignored this it was at a later moment and I assumed this error came from elsewhere when it came right from like 2 lines above all along!  fixed in 6f26447bbf0edddc05f5d89c890b134f4bb37b6a

src/targets/python/requests/client.test.ts
dimitropoulos: good feedback: did so in 6f6a18835035894b3fb5ef5a4d97b5c29761651c and made sure they were all the same while I was at it.

src/targets/shell/curl/client.ts
dimitropoulos: Screenshot_20220413_161929

This is the problem.  For better or for worse (let's be serious... FOR WORSE, haha) the curl command allowed an option of type boolean to be passed before.

src/targets/shell/curl/client.ts
dimitropoulos: I resloved it with type-fest's `Merge`: https://github.com/sindresorhus/type-fest/blob/main/source/merge.d.ts

see 7ea5e5d16bc4cad95ab9970a1b210b9a48f6c5a7

src/httpsnippet.ts
dimitropoulos: nah.  no need to apologize.  I appreciate any-and-all efforts to try to fix this.  We'll definitely need to loop back to this some time after this PR merges.

src/targets/shell/curl/client.ts
dimitropoulos: see https://github.com/Kong/httpsnippet/pull/248#discussion_r849869109

src/targets/shell/curl/client.ts
dimitropoulos: oh, for sure it's strange.. but this is how the app worked all along!

see https://github.com/Kong/httpsnippet/blob/master/src/helpers/reducer.js#L15 which is used to build up the `headersObj` object.  The values can be string arrays!!  there were even tests for this behavior https://github.com/Kong/httpsnippet/blob/master/test/reducer.js#L22.  This is a bug, in my opinion, that PRETTY MUCH all of the clients are written as though this cannot happen, when in fact it definitely can.  It's an edge case, for sure, but a real one all the same.

dimitropoulos: I added some documentation on this in 2ffb3da4a84b48cc018acc5446bf6197e0253963

filfreire (approved):

filfreire:
Awesome work @dimitropoulos

package.json
```diff
@@ -39,17 +39,16 @@
     "xmlhttprequest"
   ],
   "engines": {
-    "node": "^14.19.0 || ^16.14.0",
-    "yarn": "1.22.15"
+    "node": "^14.19.0 || ^16.14.0"
```
erunion: @dimitropoulos Want to add Node 18 in here? It hit on the 19th.

dimitropoulos: great call: did in [5be827dd7566d62ffa88787a9f01956c117635e8](https://github.com/Kong/httpsnippet/pull/248/commits/46e2bc7f27a24d5992b06052e8084a5e24f732d4)

dimitropoulos:
Here were my notes:

## TODO:
- convert to GitHub actions
- fix Dockerfile and docker-compose
- look into the prior usage of the 'debug' package
- write some CLI tests to ensure the behavior is predictable
- completely redo the README

### later:
- use prettier for js languages (and more?) but as a devDependency only to not increase bundle size
- add option to curl for minifying json

# changelog:

## Breaking
- drops support for node versions less than 14.19 (the current LTS)
- the `http` target changed its client ID to `http1.1` from `1.1`
- formatter strings (i.e. "%s" substitutions) are no longer supported (now that template strings exist, you don't need them, so please just send the string you want as the line of code as-is)
- the "dot-method-train" style API is no longer supported on the CodeBuilder.  All the same functionality exists, but just call the methods separately.  For example instead of `code.push('a').blank().push('b')`  just do `push('a'); blank(); push('b');`
- the JavaScript target will output ES6 code
- HTTPSnippet is no longer the default export (now a named export)

PR #251, opened 2022-04-26
Add Github Build Workflow (#250)

### LINKED ISSUE #250: make a simple GitHub Actions workflow
now that #248 merged, we need to implement CI.

all we need is:
- `npm install`
- `npm run test`
- `npm run lint`
- `npm run build`

all running with no errors

filfreire:
Closes #250 

Screenshot 2022-04-26 at 14 05 50

Note: The workflow will only show up, and be called for any new branches, after we merge this PR.

In order to test it out, I've forced push my branch to my fork's master branch. Here you can see example executions of this workflow: https://github.com/filfreire/httpsnippet/actions


.github/workflows/build.yml
```diff
@@ -0,0 +1,27 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout branch
+        uses: actions/checkout@v3
+
+      - name: Setup Node
+        uses: actions/setup-node@v3
+
+      - name: Install
+        run: npm install
```
DMarby: `npm install` should never be used in CI, `npm ci` should be used instead

dimitropoulos: today I learned: https://www.geeksforgeeks.org/difference-between-npm-i-and-npm-ci-in-node-js/

filfreire: @DMarby thanks 👍 
Fixed it.

filfreire: +1 TIL about `npm ci`

DMarby (approved):

.github/workflows/build.yml
```diff
@@ -0,0 +1,27 @@
...
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout branch
+        uses: actions/checkout@v3
+
+      - name: Setup Node
+        uses: actions/setup-node@v3
```
erunion, line 19: Thoughts on setting this up with a matrix to run tests across Node 14, 16, and 18?

filfreire: @erunion @dimitropoulos I just pushed this change, CI pipeline should now run for those major Node versions

filfreire:
fyi @dimitropoulos @erunion 

I'm not sure if it's intended, but the CI job [fails for Node 14](https://github.com/filfreire/httpsnippet/runs/6426474223?check_suite_focus=true#step:4:4):
```
npm ERR! fsevents not accessible from jest-haste-map
```


Seems it's a [known issue](https://stackoverflow.com/questions/68021121/fsevents-not-accessible-from-jest-haste-map), but would involve us doing some workarounds on the GH actions.

 Let me know if we should remove that and keep only support for Node 16 and 18.

dimitropoulos (commented):
I mean.... we tried.  let's just remove node 14 from the CI.  We can say node 14 (which is in maintenance mode anyway) is best-effort (which, it will be: we'll still merge things like https://github.com/Kong/httpsnippet/pull/252 as they arise).

filfreire:
> I mean.... we tried. let's just remove node 14 from the CI. We can say node 14 (which is in maintenance mode anyway) is best-effort (which, it will be: we'll still merge things like #252 as they arise).

Alright, @dimitropoulos I removed node 14 👍

PR #252, opened 2022-04-29
fix: compatibility issues on node 14 with `Object.hasOwn()`

erunion:
This resolves a compatibility issue with the new TS rewrite on Node 14 where tests, and the library, fail because [`Object.hasOwn()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/hasOwn) was introduced in Node 16.9.

Running tests on Node 14 prior to this fix:

Screen Shot 2022-04-29 at 3 02 44 PM

And after:

Screen Shot 2022-04-29 at 3 05 00 PM

changelog(Fixes): Fixed a compatibility issue with the new TS rewrite on Node 14 where tests, and the library, fail because [`Object.hasOwn()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/hasOwn) was introduced in Node 16.9

### DIFF
--- a/src/targets/php/helpers.ts
+++ b/src/targets/php/helpers.ts
@@ -25,7 +25,7 @@ export const convertType = (obj: any[] | any, indent?: string, lastIndent?: stri
     case '[object Object]': {
       const result: string[] = [];
       for (const i in obj) {
-        if (Object.hasOwn(obj, i)) {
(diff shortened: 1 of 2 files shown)


dimitropoulos (approved):
haha, and here I was all excited to be in the future and able to use Object.hasOwn().

PR #253, opened 2022-04-29
feat: addition of a PHP target for Guzzle

### LINKED ISSUE #132: PHP Guzzle generator

erunion:
> This is a replay of the work I did in https://github.com/Kong/httpsnippet/pull/228 just against the new TS rewrite.

This finishes up the work that @robertoarruda started in https://github.com/Kong/httpsnippet/pull/140 to add a new PHP target for [Guzzle](https://docs.guzzlephp.org/).

Differences between their PR and this one:

* Removed some unused/unimplemented options. 
* Fixed up all unit tests.
* Manually tested every Guzzle-generated snippet to ensure that it runs.
  * Some tweaks had to be made to multipart and cookie handling as a result of this.

Resolves #132 (and also #140).

changelog(New): adds PHP client for Guzzle

### DIFF
--- a/src/helpers/__snapshots__/utils.test.ts.snap
+++ b/src/helpers/__snapshots__/utils.test.ts.snap
@@ -231,10 +231,10 @@ Array [
         "title": "cURL",
       },
       Object {
-        "description": "PHP with guzzle v7",
+        "description": "PHP with Guzzle",
         "key": "guzzle",
         "link": "http://docs.guzzlephp.org/en/stable/",
-        "title": "Guzzle v7",
+        "title": "Guzzle",
       },
       Object {
         "description": "PHP with pecl/http v1",
--- a/src/helpers/__snapshots__/utils.test.ts.snap
+++ b/src/helpers/__snapshots__/utils.test.ts.snap
@@ -231,6 +231,12 @@ Array [
         "title": "cURL",
       },
       Object {
+        "description": "PHP with guzzle v7",
+        "key": "guzzle",
+        "link": "http://docs.guzzlephp.org/en/stable/",
+        "title": "Guzzle v7",
+      },
+      Object {
         "description": "PHP with pecl/http v1",
         "key": "http1",
         "link": "http://php.net/manual/en/book.http.php",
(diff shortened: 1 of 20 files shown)


src/helpers/__snapshots__/utils.test.ts.snap
```diff
@@ -230,6 +230,12 @@ Array [
         "link": "http://php.net/manual/en/book.curl.php",
         "title": "cURL",
       },
+      Object {
+        "description": "PHP with guzzle v7",
```
dimitropoulos: just to say it out loud, I don't have a ton of interest in keeping old versions of things around in this library as targets.

so like.. if in the future guzzle releases v8 and it's incompatible with what we have here, I'd rather we just update the guzzle target to v8 (using the same `guzzle` key) and major (or minor?) release `httpsnippet` with only that updated version.

dimitropoulos (approved):
I'm totally cool with this.  I'm not a PHP dev, but I'm assuming you are (enough, anyway, to write this).

dimitropoulos (changes requested):
Screenshot_20220513_154150

oh shoot, now we need a new fixture since https://github.com/Kong/httpsnippet/pull/258 merged.  would you mind taking a look?  I added it for you in 7203a1657dbeae0fd0a7392ab9b5de5962b288f3:

```php
<?php

$client = new \GuzzleHttp\Client();

$response = $client->request('POST', 'http://mockbin.com/har');

echo $response->getBody();
```

but it doesn't seem complete because it doesn't say anything about multipart

src/helpers/__snapshots__/utils.test.ts.snap
erunion: Yeah agreed. I'll remove the v7 bit from this.

erunion:
@dimitropoulos This is good for review again:

Screen Shot 2022-05-19 at 11 22 27 AM

erunion:
> I'm totally cool with this. I'm not a PHP dev, but I'm assuming you are (enough, anyway, to write this).

I am, or at least used to be a PHP dev, but we have this new target fully covered by code coverage within an integration suite I built in our fork. It works I swear!

dimitropoulos (approved):

dimitropoulos:
looks great

PR #254, opened 2022-04-29
fix: typo in the httpie `style` option not being correctly applied

erunion:
> This is a reworking of the original work I did on this in https://github.com/Kong/httpsnippet/pull/233 just against the new TS rewrite.

This fixes a typo, and some false-positive tests, within the `httpie` target where its `style` option wouldn't be correctly applied.

Changelog(Fixes): Fixed an issue within the `httpie` target where its `style` option wouldn't be correctly applied.


dimitropoulos (approved):
great catch

PR #255, opened 2022-04-29
fix: axios targets not sending `x-www-form-urlencoded` properly

erunion:
> This is a reworking of the my original PR for this in https://github.com/Kong/httpsnippet/pull/232 just against the new TS rewrite. 

This fixes a bug in both the Node and JS `axios` targets where they weren't sending `application/x-www-form-urlencoded` properly.

You can see this in action with the `application-form-encoded` fixture against https://httpbin.org/anything:

```js
var axios = require("axios").default;

var options = {
  method: 'POST',
  url: 'https://httpbin.org/anything',
  headers: {'content-type': 'application/x-www-form-urlencoded'},
  data: {foo: 'bar', hello: 'world'}
};

axios.request(options).then(function (response) {
  console.log(response.data);
}).catch(function (error) {
  console.error(error);
});
```

```js
{
  args: {},
  data: '',
  files: {},
  form: { '{"foo":"bar","hello":"world"}': '' },
  headers: {
    Accept: 'application/json, text/plain, */*',
    'Content-Length': '29',
    'Content-Type': 'application/x-www-form-urlencoded',
    Host: 'httpbin.org',
    'User-Agent': 'axios/0.21.4',
    'X-Amzn-Trace-Id': 'REDACTED'
  },
  json: null,
  method: 'POST',
  origin: 'REDACTED',
  url: 'https://httpbin.org/anything'
}
```

And with this fix:

```js
{
  args: {},
  data: '',
  files: {},
  form: { foo: 'bar', hello: 'world' },
  headers: {
    Accept: 'application/json, text/plain, */*',
    'Content-Length': '19',
    'Content-Type': 'application/x-www-form-urlencoded',
    Host: 'httpbin.org',
    'User-Agent': 'axios/0.21.4',
    'X-Amzn-Trace-Id': 'REDACTED'
  },
  json: null,
  method: 'POST',
  origin: 'REDACTED',
  url: 'https://httpbin.org/anything'
}
```

I've also modernized the Node `axios` target over to using `const` instead of `var` in snippets it generates.

changelog(Improvements): (internal) adds ability to add post processors to code builder

### DIFF
--- a/src/targets/javascript/axios/client.ts
+++ b/src/targets/javascript/axios/client.ts
@@ -26,7 +26,7 @@ export const axios: Client = {
       ...options,
     };
 
-    const { blank, push, join } = new CodeBuilder({ indent: opts.indent });
+    const { blank, push, join, addPostProcessor } = new CodeBuilder({ indent: opts.indent });
 
     push("import axios from 'axios';");
     blank();
@@ -55,6 +55,7 @@ export const axios: Client = {
           blank();
 
           requestOptions.data = 'encodedParams,';
+          addPostProcessor(code => code.replace(/'encodedParams,'/, 'encodedParams,'));
         }
 
         break;
@@ -103,6 +104,6 @@ export const axios: Client = {
     push('console.error(error);', 2);
     push('});', 1);
 
-    return join().replace(/'encodedParams,'/, 'encodedParams,');
+    return join();
   },
 };
--- a/src/targets/javascript/axios/client.ts
+++ b/src/targets/javascript/axios/client.ts
@@ -46,7 +46,17 @@ export const axios: Client = {
 
     switch (postData.mimeType) {
       case 'application/x-www-form-urlencoded':
-        requestOptions.data = postData.paramsObj;
+        if (postData.params) {
+          push('const encodedParams = new URLSearchParams();');
+          postData.params.forEach(param => {
+            push(`encodedParams.set('${param.name}', '${param.value}');`);
+          });
+
+          blank();
+
+          requestOptions.data = 'encodedParams,';
+        }
+
         break;
 
       case 'application/json':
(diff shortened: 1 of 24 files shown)


src/targets/javascript/axios/client.ts
```diff
@@ -89,6 +99,6 @@ export const axios: Client = {
     push('console.error(error);', 2);
     push('});', 1);
 
-    return join();
+    return join().replace(/'encodedParams,'/, 'encodedParams,');
```
dimitropoulos: it feels a little hacky and disjointed to do this replace here at the very end.  any way to do it up above?

src/targets/node/axios/client.ts
```diff
@@ -28,8 +28,7 @@ export const axios: Client = {
 
     const { blank, join, push } = new CodeBuilder({ indent: opts.indent });
 
-    push("var axios = require('axios').default;");
-    blank();
+    push("const axios = require('axios').default;");
```
dimitropoulos, line 30: happy to see ES6 starting to make it's way in here

src/targets/javascript/axios/client.ts
erunion: Unfortunately not because encodedParams in the request options needs to be a variable in the snippet, and we can only do that handling after requestoptions had been added to the snippet.

dimitropoulos: I just pushed 2343bb2736e2f23b1db718eb29a970d1c30a740d.  check it out and lemme know what you think of it.

erunion: Oh I love it. Will update this to use it.

erunion: Love it.

dimitropoulos (approved):

PR #258, opened 2022-04-29
fix: case where if `postData.params` is missing some targets crash

erunion:
A number of targets made the assumption that if you're doing a `multipart/form-data` request that you'd have `postData.params` present, and when that isn't true they'd either crash or would output useless multipart setup information in the snippet. Now normally you would have parameters present for a request like this but the [HAR spec](http://www.softwareishard.com/blog/har-12-spec/#request) specifies that `postData.params` is optional so these snippets should respect that.

changelog(Fixes): Some targets would crash if provided multipart/form-data without params


dimitropoulos (approved):
I read through all of this twice and I see nothing wrong.  good fix!  ran install/build/lint/test and all is well.

PR #259, opened 2022-04-30
feat: native upload support in python `requests` snippets

erunion:
This updates the Python target for the [requests](https://docs.python-requests.org/en/latest/) library to better natively handle file uploads with its `files` parameter, instead of supplying it an ugly and unreadable `multipart/form-data` boundary header. To help make testing this work a little easier I've expanded the `multipart-data` test fixture for every language + target to show how it showcases sending a file and a piece of form data.

changelog(Improvements): python:requests will better handle file uploads (multipart-data)


src/targets/targets.test.ts
```diff
@@ -62,7 +62,12 @@ availableTargets()
               const { convert } = new HTTPSnippet(request);
               const result = convert(targetId, clientId); //?
 
-              expect(result).toStrictEqual(expected);
+              // Some `multipart/form-data` tests construct their payload with `\r\n` line breaks,
+              // which can be problematic to save and debug within an editor so let's convert them
+              // to a standard line break instead for the sake of this assertion.
+              expect((result as string).replace(/\r\n/g, '\n').trim()).toStrictEqual(
```
dimitropoulos: would you mind please reverting this: see https://github.com/Kong/httpsnippet/pull/263#discussion_r877632362

PR #262, opened 2022-05-13
chore: add nvmrc and set it to 20

bbbco:
Add nvmrc and set it to ~~14.9~~ 18 as per https://github.com/Kong/httpsnippet/pull/248

Enables easy `nvm use` functionality.


.nvmrc
```diff
@@ -0,0 +1 @@
+14.9
```
dimitropoulos: any objection to matching it to the package.json engines.node field? https://github.com/Kong/httpsnippet/blob/62d158df9a03bfd4af89a00ec6dbb15fce848ab3/package.json#L42
```suggestion
^14.19.1 || ^16.14.2 || ^18.0.0
```

is this the right syntax for this? I don't use nvm (I use n) so I'm not sure.

jonluca: This should probably just be `18`

https://github.com/nvm-sh/nvm#nvmrc

`.nvmrc` just defines what the default version of node for this project should be. It's more along the lines of "what version of node should you be developing with", not which versions of node should this project support.

Under that definition you can either choose a) go with the most modern version, for DX and features or b) go with the earliest supported version from node modules, to catch regressions during development.

In general I'm usually in favor of A) and then letting a test suite catch regressions, or a build system to transpile newer language features down. But this is just personal preference

bbbco: A) sounds good. I'll make the change and push up shortly.

I tried signing the CLA, but the page seems broken.

filfreire (approved):

PR #263, opened 2022-05-19
fix: headers not being properly applied to R httr snippets

erunion:
This fixes a problem with R `httr` snippets where headers are being improperly added to snippets:

```r
library(httr)

url <- "https://dash.readme.com/api/v1/api-specification"

queryString <- list(
  perPage = "10"
  page = "1",
)

response <- VERB("GET", url, add_headers(x_readme-version = 'v3.0Authorization = 'Basic REDACTED'), query = queryString, content_type("application/octet-stream"), accept("application/json"))

content(response, "text")
```

* [x] Multiple headers aren't being separated with a comma
* [x] Header values aren't always being encased with a trailing apostrophe.
* [x] `x-` headers are being rewritten to `x_`.

### 🧬  Testing

I have **zero** R experience but after futzing with homebrew I was able to get it installed and this is what running the `headres.r` output fixture looks like:

Screen Shot 2022-05-19 at 11 04 42 AM

You can see `x-foo` and `x-bar` being sent as expected.

Changelog(Fixes): Fixed an issue with headers not being properly applied to R httr snippets


src/fixtures/requests/headers.json
```diff
@@ -9,6 +9,10 @@
     {
       "name": "x-foo",
       "value": "Bar"
+    },
+    {
+      "name": "x-bar",
```
erunion, line 14: Added this new header to this fixture so we can test that multiple headers are being delivered as expected.

src/targets/targets.test.ts
```diff
@@ -62,7 +62,7 @@ availableTargets()
               const { convert } = new HTTPSnippet(request);
               const result = convert(targetId, clientId); //?
 
-              expect(result).toStrictEqual(expected);
+              expect((result as string).trim()).toStrictEqual(expected.trim());
```
erunion: `.editorconfig` adding new lines but the snippet generator not has made making changes to these fixtures a bit annoying.

dimitropoulos: I know I know I know exactly what you mean.. and I did the same thing originally, but the eventual problem is that whitespace and newlines are truly significant in some cases and (as a design goal) I really want what's in these files to literally match the output - even if they do crazy things.

For example, the `http1.1` targets will get a whole lot less specific.  they even do one of the strangest things of all, which is combine CRLF and LF in the same file.... but ..... that's the spec.

would you mind reverting this?  I totally get that it's annoying, but in my view it's just a fact of life.  I do `Ctrl+P` and `save without formatting` to get around this, personally (in vscode).

dimitropoulos (changes requested):
see https://github.com/Kong/httpsnippet/pull/263#discussion_r877632362

src/targets/targets.test.ts
erunion: @dimitropoulos Done

dimitropoulos: thanks so much.  I apologize for how inconvenient it is.  always open to ideas if you can think of any that solve the issue more elegantly (is there a way, for example, to tell vscode to turn of formatters for a particular file glob?)

dimitropoulos (approved):

PR #269, opened 2022-05-27
fix: issue where query strings in R wouldn't be properly concatenated

erunion:
A user reported this issue to us on our docs where query strings for R wouldn't be properly concatenated with a comma:

```r
library(httr)

url <- "https://dash.readme.com/api/v1/categories"

queryString <- list(
  perPage = "10"    # this is missing a comma
  page = "1",       # this shouldn't have a comma
)

response <- VERB("GET", url, add_headers('Authorization' = 'Basic REDACTED'), query = queryString, content_type("application/octet-stream"))

content(response, "text")
```

Digging into it, because the `httr` target was calculating the total number of query strings before removing the `key` the count would always be off for a query string of two parameters.

Reason the existing `query` fixture tests didn't have this problem was because its query string evaluates to `{ foo: [ 'bar', 'baz' ], baz: 'abc' }`, but with a total of `3` prior to deleting `queryObj.key`.  Because 3 is greater than the number of entries it'd *actually* run through, it wouldn't ever add a comma on the last entry because it would stop running at the 2nd mark.

Super weird bug! 

Changelog(Fixes): Fixed an issue where query strings in R wouldn't be properly concatenated

### DIFF
--- a/src/targets/r/httr/client.ts
+++ b/src/targets/r/httr/client.ts
@@ -8,6 +8,11 @@
  * for any questions or issues regarding the generated code snippet, please open an issue mentioning the author.
  */
 
+export interface HttrOptions {
+  /** @default '  ' */
+  indent?: string;
+}
+
 import { CodeBuilder } from '../../../helpers/code-builder';
 import { Client } from '../../targets';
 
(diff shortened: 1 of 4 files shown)


src/targets/r/httr/client.ts
```diff
@@ -32,24 +32,22 @@ export const httr: Client = {
...
       blank();
-    } else if (queryString.length > 1) {
-      let count = 1;
-
+    } else if (queryCount > 1) {
       push('queryString <- list(');
 
-      for (const query in qs) {
-        if (count++ !== queryCount - 1) {
+      Object.keys(qs).forEach((query, i) => {
```
dimitropoulos: love it! if we could do away with every single for loop, I'd be in favor!

the only thing I'll mention is just a reminder that in JavaScript objects are unordered, so the sort order here is technically undefined (in Safari's WebKit engine, for example, it would be different than Chrome's V8).

dimitropoulos (approved):

PR #285, opened 2022-07-18
feat: Add support for insecureSkipVerify

### LINKED ISSUE #110: OpenSSL verify should not be set to NONE
https://github.com/Kong/httpsnippet/blob/1855a742777d52a74c58bbc9482539327b03d71d/src/targets/ruby/native.js#L32

This disable the check of the SSL certificate and should be prevented. Can this be changed to have SSL verified?

dimitropoulos:
closes: #238
closes: #110

This is a direct port of #238, but we can't push to it to rebase the PR because the original fork (httptoolkit) has been deleted.


sandstrom:
Great @dimitropoulos 🎉

PR #286, opened 2022-07-18
chore: minor cleanup

erunion:
* [x] Changing `let` variable instantiations in the Node Fetch client to `const`. This now matches the JS Fetch target.
* [x] Removing an unnecessary trailing space at the end of cookie declarations in Node Fetch. 
* [x] Adding missing test cases for the `addTarget` and `addTargetClient` APIs that didn't get ported over in the TS rewrite.
* [x] Finally, I found and fixed a few typos some tests.

### DIFF
--- a/src/targets/node/fetch/client.ts
+++ b/src/targets/node/fetch/client.ts
@@ -97,8 +97,8 @@ export const fetch: Client = {
     // construct cookies argument
     if (cookies.length) {
       const cookiesString = cookies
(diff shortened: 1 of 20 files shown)


src/targets/node/fetch/client.ts
```diff
@@ -107,16 +107,15 @@ export const fetch: Client = {
       }
     }
     blank();
-    push(`let url = '${url}';`);
-    blank();
```
erunion, line 111: Removed this blank line because short snippets here were uncomfortably long to read.

dimitropoulos (approved):

Issue #271, opened 2022-06-08
Quotes in header values break almost all snippets for all Chromium HARs


pimterry:
For example, I'm pretty sure this is an entirely valid raw HTTP header:

```
My-Header: "quoted" 'value'
```

In almost all cases I've seen, httpsnippet inserts this header value literally, without escaping either quote format, but within quotes in the output syntax, and so generating broken output every time, like: `headers: { "My-Header": ""quoted" 'value'" }`

This is bad because quotes are used in a lot of real HTTP traffic, for example new Chromium versions now send a "sec-ch-ua-platform" header on every single HTTPS request, and the value is always wrapped in quotes: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Sec-CH-UA-Platform. 

That means this breaks almost all snippets from almost all Chrome HAR files, making them all syntactically invalid.

The only cases I've seen where this does work is the JS snippets that use stringify-object to build an options object, which generates single-quoted output and automatically escapes any included string quotes en route.

pimterry:
I've actually been using my own [httpsnippet fork](https://github.com/httptoolkit/httpsnippet), since the repo seemed abandoned, and in my fork (based on the pre-typescript code unfortunately) @eeWynell added a %q format to the codebuilder that handles quotes automatically, which might be interesting: https://github.com/httptoolkit/httpsnippet/commit/2401900ba1fb04526d2e110cfaf15a119a024083#diff-78e5fb011bd79a6d7b6f88df468aadeac95cf9c8adc86268642815b77f8bfaf2.

That was intended to start fixing some similar issues and add a base to add other new languages, but something similar might help solve this issue here.

As an aside - I'm excited to see the many changes here recently! I'd be very happy to drop my fork and migrate back to the main package, if there's a chance that it'll be actively maintained in future. I'm using httpsnippet in [HTTP Toolkit](https://httptoolkit.tech). I'd be happy to help with the maintenance work here if there's any way I can contribute, since I'm using it heavily and currently maintaining my own fork of the same thing anyway, just let me know.

pimterry:
@filfreire @dimitropoulos @erunion Any thoughts on the above?

This breaks httpsnippet for a something like 70% of modern browser traffic right now (all Chromium-based traffic). That sec-ch-ua-platform header is widely used, it always contains quotes in the value, and the quotes are never escaped in snippets, so they all break. It'd be great to get that fixed.

pimterry:
[...]

PR #289, opened 2022-07-18
Escape quotes in headers correctly in all languages

pimterry:
As discussed in https://github.com/Kong/httpsnippet/issues/271#issuecomment-1185716335 - this ports the test changes from https://github.com/httptoolkit/httpsnippet into the new folder structure here, providing working examples for correct quote-escaping for every single language/format.

~~This does _not_ make the tests pass, we'll need to do that separately - currently this fails for 22 of the snippets, where they're currently not escaping quotes right now.~~

~~I think the base logic to fix this should look something like [this](https://github.com/httptoolkit/httpsnippet/blob/a10703297c8b03cbf62b8c5d282cbdf36596fa04/src/helpers/format.js#L3-L45), which we'll need to call in the appropriate places with the appropriate options (delimiter, escapeChar and escapeNewlines). You can see how it's used for each snippet [in the original commit](https://github.com/httptoolkit/httpsnippet/commit/a10703297c8b03cbf62b8c5d282cbdf36596fa04) to check which ones you need, but the tests here should be enough to confirm that's correct.~~

~~Happy for somebody else to look into fixing this, if anybody has time, or I should be able to migrate the fix for this myself late this week or early next if not.~~

EDIT: Now updated to add a working implementation, in addition to updating all the test fixtures to cover this.

### DIFF
--- a/src/helpers/escape.test.ts
+++ b/src/helpers/escape.test.ts
@@ -1,0 +1,35 @@
+import { escapeString } from './escape';
+
+describe('Escape methods', () => {
+  describe('escapeString', () => {
+    it('does nothing to a safe string', () => {
+      expect(
+        escapeString("hello world")
+      ).toBe("hello world");
+    });
+
+    it('escapes double quotes by default', () => {
+      expect(
+        escapeString('"hello world"')
+      ).toBe('\\"hello world\\"');
+    });
+
+    it('escapes newlines by default', () => {
+      expect(
+        escapeString('hello\r\nworld')
+      ).toBe('hello\\r\\nworld');
+    });
+
+    it('escapes backslashes', () => {
+      expect(
+        escapeString('hello\\world')
+      ).toBe('hello\\\\world');
+    });
+
+    it('escapes unrepresentable characters', () => {
+      expect(
+        escapeString('hello \u0000') // 0 = ASCII 'null' character
+      ).toBe('hello \\u0000');
+    });
+  });
+});
--- a/src/helpers/escape.ts
+++ b/src/helpers/escape.ts
@@ -1,0 +1,81 @@
+export interface EscapeOptions {
+  /**
+   * The delimiter that will be used to wrap the string (and so must be escaped
+   * when used within the string).
+   * Defaults to "
+   */
+  delimiter?: string;
+
+  /**
+   * The char to use to escape the delimiter and other special characters.
+   * Defaults to \
+   */
+  escapeChar?: string;
+
+  /**
+   * Whether newlines (\n and \r) should be escaped within the string.
+   * Defaults to true.
+   */
+  escapeNewlines?: boolean;
+}
+
+/**
+ * Escape characters within a value to make it safe to insert directly into a
+ * snippet. Takes options which define the escape requirements.
+ *
+ * This is closely based on the JSON-stringify string serialization algorithm,
+ * but generalized for other string delimiters (e.g. " or ') and different escape
+ * characters (e.g. Powershell uses `)
+ *
+ * See https://tc39.es/ecma262/multipage/structured-data.html#sec-quotejsonstring
+ * for the complete original algorithm.
+ */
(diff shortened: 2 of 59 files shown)


dimitropoulos:
wow thanks!  sounds like you can see we're catching up around here! :)

Still working on the readme, but there's a super useful shortcut I wanted to tell you about: `OVERWRITE_EVERYTHING=true npm run test`.  Like it sounds, haha, it will overwrite all tests.  It's useful when working on a change that will update fixtures.

pimterry:
All done, this now fixes #271, and there's also a tiny patch in here for weird indentation around python3 header values.

src/helpers/escape.ts
```diff
@@ -0,0 +1,81 @@
...
+ * for the complete original algorithm.
+ */
+export function escapeString(value: any, options: EscapeOptions = {}) {
+  const {
+    delimiter = '"',
+    escapeChar = '\\',
+    escapeNewlines = true
+  } = options;
+
+  value = value.toString();
```
dimitropoulos: nit: if it's easy to avoid (and it seems like it is in this case) we should avoid parameter reassignment (e.g. https://eslint.org/docs/latest/rules/no-param-reassign).  thoughts?

src/helpers/escape.ts
```diff
@@ -0,0 +1,81 @@
...
+    } else {
+      return c;
+    }
+  }).join('');
+}
+
+export const escapeForSingleQuotes = (value: any) =>
+  escapeString(value, { delimiter: "'" });
+
+export const escapeForDoubleQuotes = (value: any) =>
```
dimitropoulos, line 94: could you please add a TSDoc here explaining in layman's terms what this function does?

src/helpers/escape.ts
```diff
@@ -0,0 +1,81 @@
...
+      // using this primarily for header values that are generally (though not 100%
+      // strictly?) ASCII-only, this should almost never happen.
+      return JSON.stringify(c).slice(1, -1);
+    } else {
+      return c;
+    }
+  }).join('');
+}
+
+export const escapeForSingleQuotes = (value: any) =>
```
dimitropoulos, line 84: could you please add a TSDoc here explaining in layman's terms what this function does?

src/targets/python/python3/client.ts
```diff
@@ -65,13 +66,13 @@ export const python3: Client<Python3Options> = {
...
         if (count++ !== headerCount) {
-          push(`    '${header}': "${headers[header]}",`);
+          push(`    '${header}': "${escapeForDoubleQuotes(headers[header])}",`);
         } else {
-          push(`    '${header}': "${headers[header]}"`);
+          push(`    '${header}': "${escapeForDoubleQuotes(headers[header])}"`);
         }
       }
 
-      push('    }');
```
dimitropoulos, line 74: good catch

src/helpers/escape.test.ts
```diff
@@ -0,0 +1,35 @@
+import { escapeString } from './escape';
+
+describe('Escape methods', () => {
```
dimitropoulos, line 3: thank you for the unit tests in addition to all the fixtures!

dimitropoulos (approved):
@erunion could you please take a look?  otherwise, I'm happy with this!

@pimterry you're the first brave soul to update any of the root fixtures!  thanks!  If there is anything about the process that was needlessly difficult for you, please let me know. :)

erunion (approved):

src/helpers/escape.ts
pimterry: Sure, seems reasonable! Done.

src/helpers/escape.ts
pimterry: Done :+1:

src/helpers/escape.ts
pimterry: Done :+1:

PR #290, opened 2022-07-18
feat: add parameter to enable decode url for place holder

dimitropoulos:
closes #205

this is a direct port of #205, which we do not have access to push to.

I also added some tests

### DIFF
--- a/src/httpsnippet.test.ts
+++ b/src/httpsnippet.test.ts
@@ -217,4 +217,62 @@ describe('hTTPSnippet', () => {
       });
     });
   });
+
+  describe('options', () => {
+    describe('decodeUri', () => {
+      const input = {
+        method: 'GET',
+        url: 'http://mockbin.com/request/{id}',
+        httpVersion: 'HTTP/1.1',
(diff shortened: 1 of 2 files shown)


dimitropoulos:
upon talking it over with other maintainers, here's where we're at:

this is a very specific use-case. perhaps something we'd want, but we're not sure.  we'd really like to see more in depth use-cases on what this is useful for.  so it's not a "no" just a "not immediately right now".  the code here should be working for a while, and I added a bunch of tests for anyone to see what the behavioral change is.

Right now there are no options for the processing, and it's already the most complex part of the codebase, so it'd be useful to keep it that way if we at all can.

erunion:
We also have a similar option in our fork https://github.com/readmeio/httpsnippet/blob/main/src/index.ts#L171-L172 that allows us to control how query parameters get handled because we already encode them in HARs we specify. Maybe it'd be worth it for an overall hook system, similar to the `addPostProcessor` `CodeBuilder` method that could be hooked into at certain points in the library to control work.

PR #130, opened 2019-03-14
feat: PHP JSON body encoding

gugu:
The problem I have with existing snippets are clients, who try to concatenate JSON from example and complain about errors. I found convert function in this module and used it for JSON body encoding

before (http2 code vs curl, but I have no other examples):
```php
$body->append('{"domain":"shrt.co","links":[{"originalURL":"http://yourlongdomain.com/yourlonglink","path":"q8Wr5","title":"Some url title","tags":["string"],"allowDuplicates":false,"expiresAt":123456789000,"expiredURL":"string","password":"pa$$word","utmSource":"google","utmMedium":"cpc","utmCampaign":"spring_sale","cloaking":1,"redirectType":302}]}');

```

after 
```php
curl_setopt_array($curl, array(
  CURLOPT_URL => "https://api.short.cm/links/bulk",
  CURLOPT_RETURNTRANSFER => true,
  CURLOPT_ENCODING => "",
  CURLOPT_MAXREDIRS => 10,
  CURLOPT_TIMEOUT => 30,
  CURLOPT_HTTP_VERSION => CURL_HTTP_VERSION_1_1,
  CURLOPT_CUSTOMREQUEST => "POST",
  CURLOPT_POSTFIELDS => json_encode(array(
  'domain' => 'shrt.co',
  'links' => array(
    array(
        'originalURL' => 'http://yourlongdomain.com/yourlonglink',
        'path' => 'q8Wr5',
        'title' => 'Some url title',
        'tags' => array(
                'string'
        ),
        'allowDuplicates' => null,
        'expiresAt' => 123456789000,
        'expiredURL' => 'string',
        'password' => 'pa$$word',
        'utmSource' => 'google',
        'utmMedium' => 'cpc',
        'utmCampaign' => 'spring_sale',
        'cloaking' => 1,
        'redirectType' => 302
    )
  )
)),
  CURLOPT_HTTPHEADER => array(
    "authorization: REPLACE_KEY_VALUE",
    "content-type: application/json"
  ),
));
```

PR #291, opened 2022-07-19
feat: PHP JSON body encoding

dimitropoulos:
this is a direct port of https://github.com/Kong/httpsnippet/pull/130

The only thing not included here is the python3 and ruby/native updates, because those used the Node internal `child_process` package (https://nodejs.org/api/child_process.html) which we can't include in a library like this which is used in the browser, despite that it could work in certain node contexts.  We'd definitely be happy to see other solutions to the problem if someone would like to find a way to tackle it without calling out to Node internals.

### DIFF
--- a/src/targets/php/curl/client.ts
+++ b/src/targets/php/curl/client.ts
@@ -10,11 +10,12 @@
 
 import { CodeBuilder } from '../../../helpers/code-builder';
 import { Client } from '../../targets';
+import { convertType } from '../helpers';
 
 export interface CurlOptions {
   closingTag?: boolean;
   maxRedirects?: number;
-  nameErrors?: boolean;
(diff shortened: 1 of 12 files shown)


src/targets/php/curl/client.ts
```diff
@@ -10,11 +10,12 @@
 
 import { CodeBuilder } from '../../../helpers/code-builder';
 import { Client } from '../../targets';
+import { convertType } from '../helpers';
 
 export interface CurlOptions {
   closingTag?: boolean;
   maxRedirects?: number;
-  nameErrors?: boolean;
+  namedErrors?: boolean;
```
dimitropoulos, line 18: I verified that this was a typo

Issue #76, opened 2016-02-29
Use Ruby's Native JSON conversion rather than outputting a JSON string


nikz:
i.e instead of:

``` ruby
  body = " { \"a\": 123 }"
```

Do:

``` ruby
  require 'json'
  body = { a: 123 }.to_json
```

nikz:
cc @nickmalcolm

nickmalcolm:
:+1:

sandstrom:
A few suggestions on the ruby target:

1. ~The line `.push('http.verify_mode = OpenSSL::SSL::VERIFY_NONE') ` should be removed from the ruby target.~ Solved in https://github.com/Kong/httpsnippet/pull/285
    
    It's a huge security problem to not verify HTTPS certificates, since any MITM can swap out the certificate without any warning.
    
    https://brakemanscanner.org/docs/warning_types/ssl_verification_bypass/

2. Use JSON generate for the body, i.e.
    
    ```ruby
    body = JSON.generate({
      'key' => 123,
      'other_key' => 'my_value',
    })
    ```

3. ~Remove `require 'openssl'`, that is no longer needed in ruby 2.6 and later (2.5 is already end of life).~ (fixed in https://github.com/Kong/httpsnippet/pull/296)

dimitropoulos:
note: re 1. above, https://github.com/Kong/httpsnippet/pull/285 made the `OpenSSL::SSL::VERIFY_NONE` line configurable.

dimitropoulos:
@sandstrom I tried the approach you mentioned in https://github.com/dimitropoulos/httpsnippet/commit/46b756473d92ad8a38a43e53d822fcf709c06e69 but I'm not very knowledgeable about Ruby's syntax, or at least not knowledgeable enough to know if it's correct.  You suggested `JSON.generate` but on the web everywhere I seem to see `{}.to_json()` being used.  Any suggestions on how to proceed?

sandstrom:
@dimitropoulos Both works, but `JSON.generate` is better (more obvious, less risk of conflict with other ruby gems).

Easier to understand that this is the `json` gem being used (it's used in both cases, but with `to_json` that's not very obvious), and there is no risk of some other code having patched/shadowed the `to_json` method (unlikely, but can happen).

However, after looking at the source for Python (fairly similar to Ruby), I realize that it's  some work.

I've tried to put something together here, that would go onto these lines:

[...]

PR #296, opened 2022-07-21
removes `require 'openssl'` from ruby target (no longer needed)

dimitropoulos:
according to https://github.com/Kong/httpsnippet/issues/76#issuecomment-1006619671 we can now remove this line as of ruby 2.6


dimitropoulos:
@sandstrom could you please verify this PR is correct?

sandstrom (approved):
Looks great!

dimitropoulos:
thanks!

PR #301, opened 2022-08-31
Prep for Release: source maps, types, prettier

dimitropoulos:
We're starting to look more seriously into targeting a release.  As we do that, we noticed that the library does not have all the necessary exports.  The first commit in this PR fixes that.  While doing that, I noticed that the linting was misconfigured -> it was always autofixing.  So the next commit fixes that.  Finally, the third commit just runs that autofixer and commits it.


marckong (approved):
LGTM!

filfreire (approved):

PR #323, opened 2023-07-18
Fix setting default value of request.PostData of HarEntry

nsrCodes:
since har entry is of type [[src](https://github.com/Kong/httpsnippet/blob/master/src/httpsnippet.ts#L36-L54)] 
```
interface Entry {
  request: Partial<HarRequest>;
}
```

so `request.postData` could have the value `undefined`

The [current approach](https://github.com/Kong/httpsnippet/blob/master/src/httpsnippet.ts#L103-L107) of setting the default value for this key will fail when `postData` is `undefined`

This then breaks the [switch case inside the prepare method](https://github.com/Kong/httpsnippet/blob/master/src/httpsnippet.ts#L167) because there is no optional chaining

This PR is broken into two commits. If only either one feels sufficient, the changes can be cherry-picked or both can be squashed.

### DIFF
--- a/src/httpsnippet.ts
+++ b/src/httpsnippet.ts
@@ -164,7 +164,7 @@ export class HTTPSnippet {
       request.allHeaders.cookie = cookies.join('; ');
     }
 
-    switch (request.postData.mimeType) {
+    switch (request?.postData.mimeType) {
       case 'multipart/mixed':
       case 'multipart/related':
       case 'multipart/form-data':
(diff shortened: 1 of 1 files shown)


src/httpsnippet.ts
```diff
@@ -164,7 +164,7 @@ export class HTTPSnippet {
       request.allHeaders.cookie = cookies.join('; ');
     }
 
-    switch (request.postData.mimeType) {
+    switch (request?.postData?.mimeType) {
```
erunion: Not sure why this would be optionally chained because `mimeType` is a required part of `postData`.

nsrCodes: yes this wasn't required, I have updated the last commit

filfreire (approved):

PR #349, opened 2024-05-16
fix(CI): add pat for triggering releases [no-ticket]

filfreire:
adds a release PAT secret to trigger the SAST workflow from this one https://docs.github.com/en/actions/using-workflows/triggering-a-workflow#triggering-a-workflow-from-a-workflow


saisatishkarra (approved):
gtm

PR #353, opened 2024-05-17
fix: rm extra commit on release.yml [INS-3853]

filfreire:
npm version already creates a commit for us


saisatishkarra (commented):
Gtm

saisatishkarra (approved):

Issue #312, opened 2023-02-15
C# RestSharp is using deprecated syntax


jon-batscha:
The example generated for [this example](https://apiembed.com/) is currently broken for folks using versions >= v107 of RestSharp. I have also seen the RestSharp generator break similarly for every other example i have tried, both directly in httpsnippet, as well as embedded in platforms.

This is due to [breaking changes](https://restsharp.dev/v107/#restsharp-v107) released in that version.

Is anyone currently working on a fix for this? Would love to connect.

Cheers!
Jon

erunion:
@jon-batscha Would you be willing to submit a PR to fix this?

llimllib:
@jon-batscha an updated version has been merged into `readmeio/httpsnippet`, would love to know if it works for you

PR #366, opened 2024-09-15
correct RestSharp snippets (return type and Method enum)

### LINKED ISSUE #365: CSharp RestSharp snippets don't generate valid members of `RestSharp.Method` enum
Here's a portion of the code generated from a csharp RestSharp code snippet

```csharp
var request = new RestRequest(Method.GET);
```

The issue is that the actual member we need is `Method.Get` as shown in the source code here:

https://github.com/restsharp/RestSharp/blob/0fba5e727d241b1867bd71efc912594075c2934b/src/RestSharp/Enum.cs#L70

None of the Http Verbs are upper-cased in the enum definition.

Seems like I relatively easy fix. I'll try to write a PR.

michaelgwelch:
Ensure that code snippets use enum members that are pascal cased and not upper cased. For example use `Method.Get` instead of `Method.GET` to ensure the snippets can compile.

Also removes usage of unknown type `IRestResponse`

fixes #365 #367 #312

### DIFF
--- a/src/targets/csharp/restsharp/client.ts
+++ b/src/targets/csharp/restsharp/client.ts
@@ -20,8 +20,15 @@ export const restsharp: Client = {
       return 'Method not supported';
     }
 
+    function toPascalCase(str: string): string {
+      return str.replace(
+        /\w+/g,
+        word => word.charAt(0).toUpperCase() + word.slice(1).toLowerCase(),
+      );
+    }
+
     push(`var client = new RestClient("${fullUrl}");`);
-    push(`var request = new RestRequest(Method.${method.toUpperCase()});`);
+    push(`var request = new RestRequest("", Method.${toPascalCase(method)});`);
 
     // Add headers, including the cookies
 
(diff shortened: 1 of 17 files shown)


src/targets/csharp/restsharp/client.ts
```diff
@@ -20,8 +20,15 @@ export const restsharp: Client = {
...
+    function toPascalCase(str: string): string {
+      return str.replace(
+        /\w+/g,
+        word => word.charAt(0).toUpperCase() + word.slice(1).toLowerCase(),
+      );
+    }
+
     push(`var client = new RestClient("${fullUrl}");`);
-    push(`var request = new RestRequest(Method.${method.toUpperCase()});`);
+    push(`var request = new RestRequest(Method.${toPascalCase(method)});`);
```
michaelgwelch: I jumped the gun on this. This also doesn't work because `RestRequest` expects a string not an enum. At a minimum I'll need a `ToString()`. I'll need to check the RestSharp docs. I'm sure they have a recommended practice.

michaelgwelch: This should be 

* Initialize the `RestClient` with a base url
* Initialize the `RestRequest` with two parameters: relative url and `Method`.

michaelgwelch: Unfortunately I don't see that the `request` preserves base url and relative url. So I guess we can just initialize the client with full url and pass `null` or `string.empty` as the relative url but that doesn't really show a consumer the right way to do things.

filfreire (approved):

Issue #335, opened 2024-01-10
Error: strict mode: unknown keyword: "afterRequest"


Harsh062:
I am getting the below error when executing the basic example provided in the docs in my NodeJs project:

```
/Users/harsh/Desktop/http-snippet-demo/node_modules/ajv/dist/compile/index.js:120
        throw e;
        ^

Error: strict mode: unknown keyword: "afterRequest"
    at checkStrictMode (/Users/harsh/Desktop/http-snippet-demo/node_modules/ajv/dist/compile/util.js:174:15)
    at checkUnknownRules (/Users/harsh/Desktop/http-snippet-demo/node_modules/ajv/dist/compile/util.js:32:13)
    at checkKeywords (/Users/harsh/Desktop/http-snippet-demo/node_modules/ajv/dist/compile/validate/index.js:120:34)
    at validateFunctionCode (/Users/harsh/Desktop/http-snippet-demo/node_modules/ajv/dist/compile/validate/index.js:19:9)
    at Ajv.compileSchema (/Users/harsh/Desktop/http-snippet-demo/node_modules/ajv/dist/compile/index.js:80:45)
    at Ajv.resolveSchema (/Users/harsh/Desktop/http-snippet-demo/node_modules/ajv/dist/compile/index.js:193:23)
    at Ajv.getSchema (/Users/harsh/Desktop/http-snippet-demo/node_modules/ajv/dist/core.js:279:43)
    at validateHarRequest (/Users/harsh/Desktop/http-snippet-demo/node_modules/httpsnippet/dist/helpers/har-validator.js:66:24)
    at /Users/harsh/Desktop/http-snippet-demo/node_modules/httpsnippet/dist/httpsnippet.js:262:56
    at Array.forEach (<anonymous>)
```

Example Code:
```
const express = require('express');
const { HTTPSnippet } = require('httpsnippet');

const app = express();

const snippet = new HTTPSnippet({
  method: 'GET',
  url: 'http://mockbin.com/request',
});

const options = { indent: '\t' };
const output = snippet.convert('shell', 'curl', options);
console.log(output);

const server = app.listen(8000, '0.0.0.0', () => {
  console.log('Example app listening at http://%s:%s', server.address().address, server.address().port);
});
```
httpsnippet version in package.json file: **^3.0.1**
NodeJs version: **18.17**

I am getting the same error via CLI as well when executing the below command:
`httpsnippet example.json --target shell --client curl --output ./examples`

[...]

Issue #270, opened 2022-05-30
Pre-compile HAR schema for validation


captn3m0:
`Ajv` is currently used in `httpsnippet` for [validating the HAR schema passed](https://github.com/Kong/httpsnippet/blob/f64080d7d5057b2df9d4223839e001ca8009afa5/src/helpers/har-validator.ts#L21). Using complete Ajv dependency makes `httpsnippet` require `'unsafe-eval` in the CSP policy, which is a security concern. (https://github.com/ajv-validator/ajv/blob/c3e203c2696e42ec107fe6a5a659fca54b2e4993/docs/security.md#content-security-policy).

As a workaround, since the schema for HAR is known (and already included), it can be pre-compiled into a validation function as part of the build. The [Ajv Docs](https://github.com/ajv-validator/ajv-cli#compile-schemas) have details on the same. 

This would move Ajv from a runtime dependency to a optional developer-dependency (as Ajv-cli) - only needed when `har-schema` is updated. This would bring a lot more benefits, including performance and bundle size improvements.

This request comes [via the Stoplight Elements](https://github.com/stoplightio/elements/issues/2141) project, which uses `httpsnippet` internally. Since httpsnippet doesn't offer a way to pre-compile schemas, this ends up breaking Elements in configurations which limit `unsafe-eval` via CSP.

jonluca:
I did exactly this and published the library to https://npmjs.com/package/har-validator-compiled, should help out with a few of the places that use har-scheme validation.

I just put up a diff to move over to this validator. https://github.com/Kong/httpsnippet/pull/297

PR #371, opened 2024-09-30
feat: replace ajv usage [INS-145]

### LINKED ISSUE #360: Httpsnippet uses ajv but doesn't actually depend on it
In an empty directory:

```bash
$ npm install httpsnippet
$ node
> require('httpsnippet')
Uncaught Error: Cannot find module 'ajv'
```

Used here: https://github.com/Kong/httpsnippet/blob/63650f3f6a921e4a617316ba12f14904d9af4b76/src/helpers/har-validator.ts#L1

Ajv is a dev dependency in this repo (via eslint) but not an actual dependency of the project.

vaibhavrajsingh2001:
The original aim was to replace the `ajv` JSON schema parser with a parser that's compatible with Cloudflare workers.
But turns out we don't really need to use a JSON schema parser.
We were only using it to validate HAR request schema during runtime, but we can instead use pre-compiled schema validators, like [`har-validator-compiled`](https://www.npmjs.com/package/har-validator-compiled).

Resolves: https://github.com/Kong/httpsnippet/issues/360 https://github.com/Kong/httpsnippet/issues/270 https://github.com/Kong/httpsnippet/issues/335


vaibhavrajsingh2001:
This was attempted previously as part of https://github.com/Kong/httpsnippet/pull/297 too.

jackkav (approved):

PR #380, opened 2025-07-14
fix R code for special character query parameters

### LINKED ISSUE #379: Bug: R Code Generation Does Not Quote Parameters with Special Characters (e.g., `codes[]`)
When using the library to convert an HTTP request into an R code snippet, **parameters with special characters** (such as `codes[]` or `demo-param`) are not being quoted correctly in the generated R code. This causes an error in R because such parameters are not valid without quotes.

In R, certain special characters like `[]`, `-`, and others are not valid in unquoted object names. As a result, the generated R code fails to execute properly when using such parameters.

I also saw a test case at "https://github.com/Kong/httpsnippet/blob/master/src/targets/r/httr/fixtures/nested.r" which doesn't run in R,
it gives following error
```
Error: unexpected '=' in:
"queryString <- list(
  foo[bar] ="
Execution halted
```

Instead it should be like this to run correctly and avoid any special character issues.
```
queryString <- list(
  `foo[bar]` = "baz,zap",
  `fiz` = "buz"
)
```

abansal21:
Resolves #379

### DIFF
--- a/src/targets/r/httr/client.ts
+++ b/src/targets/r/httr/client.ts
@@ -48,7 +48,7 @@ export const httr: Client = {
 
     if (entriesCount === 1) {
       const entry = entries[0];
-      push(`queryString <- list(${entry[0]} = "${entry[1]}")`);
+      push(`queryString <- list(\`${entry[0]}\` = "${entry[1]}")`);
       blank();
     } else if (entriesCount > 1) {
       push('queryString <- list(');
@@ -56,7 +56,7 @@ export const httr: Client = {
       entries.forEach(([key, value], i) => {
         const isLastItem = i !== entriesCount - 1;
         const maybeComma = isLastItem ? ',' : '';
-        push(`${key} = "${value}"${maybeComma}`, 1);
+        push(`\`${key}\` = "${value}"${maybeComma}`, 1);
       });
 
       push(')');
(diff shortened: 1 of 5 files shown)


erunion (approved):

abansal21:
After reviewing this I suggest that we go with removing queryString altogether and just use url,

for an example, take GET request to "https://postman-echo.com/get?codes[web]=1&codes=2&test[]=2&codes=1"

it generates
```r
library(httr)

url <- "https://postman-echo.com/get"

queryString <- list(
  `codes[web]` = "1",
  `codes` = "2,1",
  `test[]` = "2"
)

response <- VERB("GET", url, query = queryString, content_type(""))

content(response, "text")
```

which gives a response
<truncated to show only args>
```json
{
  "args":{
    "codes":{
      "web":"1",
      "2,1":true
    },
    "test":["2"]
  }
. . .
}
```
But the actual response should be 
```json
"args": {
    "codes": {
      "0": "2",
      "1": "1",
      "web": "1"
    },
    "test": [
      "2"
    ]
  }
...
```
you can see this just by opening the above link

So the correct code generated should be

```r

library(httr)

url <- "https://postman-echo.com/get?codes[web]=1&codes=2&test[]=2&codes=1"

response <- VERB("GET", url, content_type(""))

content(response, "text")
```

I have this change and can make it in the same pr if the maintainers believe that its appropriate.

PR #390, opened 2026-03-25
chore: bump deps to fix vulns

### LINKED ISSUE #374: Can we have the new releases on npm?
Hi! ~~Would it be possible to get the latest release on npm?~~ Would love to use the rust target!

Edit: could the last release be set as latest on NPM?

### LINKED ISSUE #377: Node 22 Support
There _probably_ won't be any issue if we added node 22 to the list of engines supported and node versions for github actions, so if it's reasonably stress free then we should do it.

ryan-willis:
Drops the engine restriction to unblock bumping in [Kong/insomnia](https://github.com/Kong/insomnia)
Bumps the GHA step versions
Adds Node versions 20, 22, and 24 to the actions

closes #374
closes #377


.github/workflows/build.yml
```diff
@@ -39,13 +39,13 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        node-version: [16, 18, 20]
+        node-version: [16, 18, 20, 22, 24]
```
shelby-moore: Do we want to drop the older versions?

shelby-moore (approved):

.github/workflows/build.yml
ryan-willis: ooo ya, 18 is EoL and 20 is close

ryan-willis: also dropped the 16/18 as required PR checks

